PR #8, opened 2015-09-05
WIP Hub widget: initial commit

janezd:
As advertised in https://docs.google.com/document/d/1ulaWDXqB6RFPOOFvaSc8cnOs0yLwVYfMYrKov4ueUi0/edit?usp=sharing

### DIFF
--- a/orangecontrib/prototypes/widgets/owhub.py
+++ b/orangecontrib/prototypes/widgets/owhub.py
@@ -1,0 +1,23 @@
+from PyQt4.QtGui import QSizePolicy
+from Orange.widgets import widget, gui
+
+
+class OWHub(widget.OWWidget):
+    name = "Hub"
+    description = "Universal hub widget"
+    icon = "icons/Hub.svg"
+
+    inputs = [("Object", object, "get_input", widget.Default)]
+    outputs = [("Object", object, widget.Dynamic)]
+
+    NOTHING = "Nothing on input"
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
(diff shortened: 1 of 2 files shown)


orangecontrib/prototypes/widgets/owhub.py
```diff
@@ -0,0 +1,23 @@
...
+    description = "Universal hub widget"
+    icon = "icons/Hub.svg"
+
+    inputs = [("Object", object, "get_input", widget.Default)]
+    outputs = [("Object", object, widget.Dynamic)]
+
+    NOTHING = "Nothing on input"
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
```
astaric: OWWidget's init does not have parameters in Orange3.
https://github.com/biolab/orange3/blob/master/Orange/widgets/widget.py#L142

Setting parent (by canvas) is handled in widget's __new__ method.

kernc:
That spec text is excerpt from a future advertorial blog post, I hope.

orangecontrib/prototypes/widgets/owhub.py
janezd: Fixed. I actually copied this from somewhere; it turns out many widgets still have arguments - some even signalManager - in __init__

janezd:
The text is for blog, if we decide so. The whole widget is a bit of a joke, but it is also quite practical (i.e. practical joke :). In lectures I would need it exactly for the purpose described in the "advertisement" (top part, the bottom is not serious).

PR #10, opened 2015-09-07
Hotplug network add-on.

astaric:
Hotfix for everyone that uses Orange before  https://github.com/biolab/orange3/pull/693 was merged.

Network will show in the add-ons dialog after the new version of prototype is installed.


kernc:
Is this the correct repository?

astaric:
While pypi search is broken, we need this https://github.com/biolab/orange3/pull/671 to install orange3-network add-on. But in order to get that commit, all attendees would need to reinstall Orange, which is painful. By merging this, we achieve the same thing by just reinstalling the prototypes addon, which will monkeypatch the OFFICIAL_ADDONS list. Magic.

After the workshop is done, we definitely want to revert this, but I have created the pull request as an option we (@janezd) can pick.

PR #29, opened 2016-01-29
owsilhouetteplot: Warn/error on discrete variables in the input

ales-erjavec:


orangecontrib/prototypes/widgets/owsilhouetteplot.py
```diff
@@ -108,21 +108,33 @@ def set_data(self, data):
...
-                    self.cluster_var_idx = 0
+
+        if data is not None:
+            ncont = sum(v.is_continuous for v in data.domain.attributes)
+            ndiscrete = len(data.domain.attributes) - ncont
+            if ncont == 0:
+                data = None
+                error_msg = "No continuous columns"
+            elif ncont < len(data.domain.attributes):
+                warning_msg = "{0} discrete columns will not be used for " \
```
kernc: Could you condition this check simply on `data.domain.has_discrete`?

ales-erjavec: No. `AttributeError: 'Domain' object has no attribute 'has_discrete'` (biolab/orange3#978 is not yet merged)

kernc: If it eventually does get merged, won't it be easier to replace [its equivalent](https://github.com/biolab/orange3/blob/master/Orange/data/domain.py#L267)?

PR #33, opened 2016-04-15
[WIP] Pythagoras tree widget

pavlin-policar:
Pythagoras tree widget for visualizing trees.
_As requested, I've moved the PR to this repo._


BlazZupan:
Traceback (most recent call last):
  File "owclassificationpythagorastree.py", line 111, in <module>
    main()
  File "owclassificationpythagorastree.py", line 96, in main
    ow = OWClassificationPythagorasTree()
  File "/Users/blaz/dev/orange3-prototypes/orangecontrib/prototypes/widgets/owpythagorastree.py", line 87, in __init__
    self.inline_graph_report()
AttributeError: 'OWClassificationPythagorasTree' object has no attribute 'inline_graph_report'

janezd:
@pavlin-policar, buttons are now placed horizontally by default, so the method `inline_graph_report` has been removed. Just delete the call.

ajdapretnar:
Tested it and it looks nice. Just a few comments.
Classification Pythagoras Tree could probably have a more snug visualization (too much white space on top and bottom).
Also would be nice if CPT would accept color scale from Color widget. (Regression widget works well with it.)
To-do: implement Save Graph and Report for both. (just a reminder)

pavlin-policar:
I've implemented support for the Color widget for the CPT, but I haven't been able to figure out how I could implement this for the RPT, so I've left the colors for the regression widget as is.

janezd:
What kind of problem do you have with colors for regression? If you don't know how to construct to color:
first you need to create the palette with something like

    palette = ContinuousPaletteGenerator(*variable.colors)

where `variable` is the continuous variable that defines the coloring. The color is then

    palette.getRGB(value)

where `value` is the value of the `variable`. `values` can also be a numpy array, if that's more convenient.

pavlin-policar:
I don't have problems with the regression colors themselves, just with how the Color widget affects them and where I could actually use that in the widget. Getting the colors themselves isn't a problem - my current implementation works the way you described.

The problem is that I don't understand what the Color widget changes on the dataset with continuous variables, because for discrete variables it's very straightforward - it just adds `domain.class_var.colors` to the dataset, which is very easy to use.

I am not even sure if I could find a good way to use the colors for the continuous variables in my tree since I'm coloring the tree the same way as the existing tree viewers (which is strange), but having the ability to play around with it would be nice.

PR #35, opened 2016-06-20
[WIP] Pythagorean forest widget

pavlin-policar:
Visualise a random forest through a grid of pythagoras trees, which can be selected and inspected further by a tree visualisation widget.


BlazZupan:
I am moving this in. Pythagorean trees and forest will move to the core Orange (Visualization pane). Until then, these great widgets should be available in Prototypes.

PR #38, opened 2016-08-29
Add a face detection widget based on OpenCV.

thocevar:


ajdapretnar:
You probably need to add an OpenCV dependency to requirements.txt. :) Otherwise look ok to me, only that I somewhat suspicious results (it's probably an issue of embeddings, so don't worry).

ajdapretnar:
Oh, there's no requirements.txt in the first place. Anyway, widget is not working unless you manually install opencv.

kernc:
OpenCV is _not_ pip-installable.

ajdapretnar:
Yes, we'll sort it locally, for now.

PR #40, opened 2016-09-05
OWNeighbours: Output sorted neighbours

VesnaT:


orangecontrib/prototypes/widgets/owneighbours.py
```diff
@@ -86,16 +86,16 @@ def apply(self):
...
-        l = list(np.argsort(new.flatten()))[::-1]
-        s = set()
-        while len(l) > 0 and len(s) < self.n_neighbors:
-            i = int(l.pop() / len(self.reference))
-            if self.data[i] not in self.reference or not self.exclude_reference:
-                s.add(i)
-        print(distance.name, s)
-        neighbours = self.data[list(s)]
+        dist = distance(np.vstack((self.data, self.reference)))[:len(self.data),
+               len(self.data):]
```
kernc: What indentation style is this?

orangecontrib/prototypes/widgets/owneighbours.py
```diff
@@ -86,16 +86,16 @@ def apply(self):
...
+        sorted_indices = list(np.argsort(dist.flatten()))[::-1]
```
kernc:
```py
sorted_indices = list(np.argsort(dist.ravel())[::-1])
```
It's much faster to reverse an array than a list. Also use ravel() when copy is not _required_.

ajdapretnar:
@VesnaT I'm having some problems with the workflow. @kernc has made a new widget 'Webcam Capture' and it's sending new data from the webcam. Neighbors doesn't seem to recognize this new input   (it comes in as 'Reference'). My guess is it's a problem of domain matching. Could you please check? Thanks!

ajdapretnar:
As a comment: We discussed the above 'problem' with Vesna and realized it's not a problem at all, it's how the widget works. Thus I merged the PR.

PR #41, opened 2016-09-06
Webcam Capture widget

kernc:


ajdapretnar:
Seems ok to me. However:
1.) There's still an issue of wrong coloring, which is hilarious (see Slack).
2.) I just realized we probably need some quick shortcut to clear the 'Orange user' field, when we get new people in. Simple 'New User' button or 'X' should suffice.
3.) When I open the widget, the initial camera image is super-small. I think we should set minimum width-height.

The rest is ok. Icon perfectly matches my own camera, how did you know? :smile:

kernc:
1. Fixed, added Avatar filter checkbox
2. Select all on focus. Reset/Clear buttons are usability no-gos.
3. Fixed. I didn't manage to center the QLabel, though. Ideas welcome.

ajdapretnar:
Ad 2. Now a new user has to type over someone else's name (old user). It's very inconvenient in my opinion. What are proper solutions for that?

kernc:
What do you mean, "has to type over someone's name"? They can press Backspace or Delete key if they are as incompetent as my grandparents, otherwise on any first key press, the field will clear, replacing the contents with that first letter.

kernc:
Updated.

ajdapretnar:
> on any first key press, the field will clear

It doesn't clear. The text is only highlighted, which means pressing Backspace again to get empty field. I don't feel good having other people's name there, so demonstrators will have to delete this over and over again. I'm just trying to make it easy with a click of a button.

kernc:
Yes, or you can just focus the field and start typing the new name (saves one keypress). There are no privacy concerns. Orange is a single-user app.

janezd:
This widget will be used in public shows. A user (say a child) will come, take a picture, input his name and leave it there.

I discussed it with @BlazZupan and we suggest removing the name field altogether. It is unnecessary and may cause privacy issues.

The alternative is to put the name next to the capture button and clear the name immediately after capturing image.

kernc:
> The alternative is to put the name next to the capture button and clear the name immediately after capturing image.

:+1:

ajdapretnar:
> I discussed it with @BlazZupan and we suggest removing the name field altogether.

+1 for this. I suggest going with 'Me' as default as this will also be shown in the Lookalike widget. Sorry for complications on this one.

Another thing. As mentioned yesterday, Neighbors is not finding our selfie as its own closest neighbor. I suspect mismatching of domain to be an issue (however, only metas are not matching, so it's quite weird). Or perhaps it's a problem of webcam outputting .png, when all others are .jpg. I don't know, but since two widgets that normally work well don't work with Webcam, perhaps it's worth looking at issues on this side. Could we experiment with matching the domain properly and exporting .jpg? See what happens? I don't know, I accept other ideas, too. :)

ajdapretnar:
Ok. Just a quick sum up. Neighbors works fine. Face Detection sometimes has problems, but there's nothing we can do about it. Webcam is ok, expect that I'd definitely remove Name field altogether. Also, is it possible to adjust the aspect ratio of the webcam? To make it 4:5 as in our core data set? It's nothing major and we can do without, but it might be nice to have two similar photos in terms of AR. That's it, out and out.

kernc:
What's wrong with 'One Happy Orange'?

The widget outputs a second signal that contains a 4:5 image.

ajdapretnar:
> What's wrong with 'One Happy Orange'?

Nothing per se, but it might be a bit too much. :smile: Also we will have 'Me' as default in Lookalike widget where the two pictures will be compared, so I vote for 'Me'.

I'll test and merge asap!

ajdapretnar:
4:5 works great for me!

Can I have just a quick input from @astaric regarding the 'One Happy Orange' default and the Name field vs. no name field and 'Me' as default?

kernc:
So if Lookalike is overriding this anyway, we can just as well make this widget generic. Changed "Name" field into "Title", changed default title into "Snapshot {timestamp}".

ajdapretnar:
Don't you think Snapshot 1934949... is a bit ugly? I mean, I don't entirely object to it and making a widget more widely applicable is a huge pro, but from what I know this widget outputs only one image (in every case), so timestamp in image name seems a bit redundant. Would be better to put timestamp as a datetime attribute, should one want to analyse selfies in time. :grimacing:

kernc:
Done.

PR #42, opened 2016-09-06
Removed Polynomial Regression widget 

PrimozGodec:
This widget was moved to Educational add-on


lanzagar:
Maybe expand the replaces list in the new one with this version from prototypes as well.

PrimozGodec:
It is done with https://github.com/biolab/orange3-educational/pull/21

PR #44, opened 2016-09-12
OWNeighbours: Add percentage of similarity to metas

VesnaT:


ajdapretnar:
I have a huge problem. This PR works great with our demo workflow. But aren't we planning to one day put Neighbours into core Orange? In this case we would't want to have such similarity measures appended... I have no idea where else to put it or how to deal with this properly....

kernc:
>  In this case we wouldn't want to have such similarity measures appended

Why not?

ajdapretnar:
Because this is made solely for the purpose of demo events. As discussed with @VesnaT, we will solve such issues when we will move this to core Orange. Until then, since it's in Prototypes, this should be ok. Merging.

kernc:
If a widget gives me a list of neighbors for a certain example, I don't at all mind knowing how close or distant those neighbors are.

PR #47, opened 2016-09-27
OWWebcamCapture: try capture devices in LIFO order

kernc:


ajdapretnar:
Works well for me. However, smaller screens evidently have issues with HD, so I cannot resize Webcam Capture. The idea was to set Webcam and Lookalike side-by-side. @VesnaT with your recent merge, I am able now to resize Lookalike and it works well on larger screen, yet when I tested this on my laptop with HD cam, things aren't as pretty. 😞 Ideas?

kernc:
Please try again.

ajdapretnar:
Works well on Windows, still having issues on Linux, but I presume it's just my Linux incompetence.

kernc:
Resizing that QLabel was indeed something ...

How about now?

PR #49, opened 2016-10-04
[WIP] Merge widgets with different input types, RFC

pavlin-policar:
A proposal on how to merge widgets together when they take multiple input types and how to separate out their handlers. This proposal uses a mixin and decorators to register different handler types to individual classes. While the implementation covers only `Orange.data.Table` type inputs, this will be expanded to all input types, so it will be able to be used in any widget that will need to handle multiple input types.
A toy example on how using the mixin should look:

```python
class Widget(MultiInputMixin):
    my_property = 1
    attribute = 'none'

    def set_data(self, data):
        pass

    def dummy_method(self):
        pass


@Widget.data_handler(target_type=InputTypes.DISCRETE)
class Discrete:
    attribute = 'discrete'

    def simple_method(self, *args):
        pass


@Widget.data_handler(target_type=InputTypes.CONTINUOUS)
class Continuous:
    attribute = 'continuous'

    def simple_method(self, *args):
        pass

```

A full example can be seen [here](https://github.com/pavlin-policar/orange3-prototypes/blob/cr-merge-widgets/orangecontrib/prototypes/widgets/owknn.py#L15-L76).

Also, since the widget gui needs to be updated in almost all cases when the widget receives new data, it might also be beneficial to add another decorator e.g. `@update_on_input` that would be added to methods that update the gui, and would automatically be called when the widget receives new data. This was, we would avoid the need to perform all updates in `set_data`, so it doesn't have to worry about the gui.

Pros:
 - All the methods and properties that handle a single input type are grouped together in the same class, which should make understanding and debugging fairly simple.
 - The very minimal API should make converting existing widgets that are separated to a single widget fairly simple.

Cons:
 - Pylint will undoubtedly complain in the handler classes since handlers will probably need to access properties of the base class, but since the handler methods are bound to the base class at runtime, they technically don't exist in the handler, hence the errors.

The implementation is not yet functional on orange widgets, there are still several issues I need to work out, but any comments or suggestions on the API would help a great deal.

### DIFF
--- a/orangecontrib/prototypes/widgets/owknn.py
+++ b/orangecontrib/prototypes/widgets/owknn.py
@@ -1,0 +1,89 @@
+from Orange.regression import KNNRegressionLearner
+from PyQt4.QtCore import Qt
+
+from Orange.data import Table
+from Orange.classification import KNNLearner
+from Orange.widgets import gui
+from Orange.widgets.settings import Setting
+from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
+from orangecontrib.prototypes.widgets.utils.multiinput import (
+    MultiInputMixin,
+    InputTypes,
+)
+
+
+class OWKNNLearner(OWBaseLearner, MultiInputMixin):
+    name = "Nearest Neighbors"
+    description = "Predict according to the nearest training instances."
+    icon = "icons/KNN.svg"
+    priority = 20
+
+    LEARNER = KNNLearner
+
+    # TODO The MultiInput metaclass should take care of this
+    handlers = {}
+    trigger = 'set_data'
+
+    weights = ["uniform", "distance"]
+    metrics = ["euclidean", "manhattan", "chebyshev", "mahalanobis"]
+
+    learner_name = Setting("kNN")
+    n_neighbors = Setting(5)
+    metric_index = Setting(0)
+    weight_type = Setting(0)
+
+    def __init__(self):
+        super().__init__()
+        MultiInputMixin.__init__(self)
+
+    def add_main_layout(self):
+        box = gui.vBox(self.controlArea, "Neighbors")
+        self.n_neighbors_spin = gui.spin(
+            box, self, "n_neighbors", 1, 100, label="Number of neighbors:",
+            alignment=Qt.AlignRight, callback=self.settings_changed)
+        self.metrics_combo = gui.comboBox(
+            box, self, "metric_index", orientation=Qt.Horizontal,
+            label="Metric:", items=[i.capitalize() for i in self.metrics],
+            callback=self.settings_changed)
+        self.weights_combo = gui.comboBox(
+            box, self, "weight_type", orientation=Qt.Horizontal,
+            label="Weight:", items=[i.capitalize() for i in self.weights],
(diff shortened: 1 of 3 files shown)


orangecontrib/prototypes/widgets/owknn.py
```diff
@@ -0,0 +1,89 @@
...
+@OWKNNLearner.data_handler(target_type=InputTypes.DISCRETE)
+class KNNClassification:
+    LEARNER = KNNLearner
+    weights = ["uniform"]
+
+
+@OWKNNLearner.data_handler(target_type=InputTypes.CONTINUOUS)
+class KNNRegression:
+    LEARNER = KNNRegressionLearner
+    weights = ["distance"]
```
kernc: Both problem types can use either uniform or distance-based voting.

kernc:
This seems excessive. Is the functionality of classification vs. their regression widget counterparts so vastly different? Couldn't most widgets just set:
```py
if is_classification_problem(data):
    self.learner = KNNLearner
    self.learner_params = something
else:
    self.learner = KNNRegressionLearner
    self.learner_params = something_else
```
in their `_set_data()` method?

Or even _OWBaseLearner could pick the right learner like:
```py
self.learner = self.LEARNERS[is_classification_problem(data)]
```
or something simple like that??

orangecontrib/prototypes/widgets/owknn.py
pavlin-policar: This is in there just for testing purposes so the changes are visible on the actual widget (when I actually get it fully working).

pavlin-policar:
I agree, the benefits aren't visible at all on the KNN learner widget, since the only thing overridden is the `LEARNER` property. I chose KNN because I wanted to start simple.

If you take a look at the [OWSVMRegression](https://github.com/biolab/orange3/blob/master/Orange/widgets/regression/owsvmregression.py), you can see a lot more is overridden. Even further, the [OWLinearRegression](https://github.com/biolab/orange3/blob/master/Orange/widgets/regression/owlinearregression.py) and [OWLogisticRegression](https://github.com/biolab/orange3/blob/master/Orange/widgets/classify/owlogisticregression.py) widgets are to be merged, which, at first glance, appear to have even less in common.

Also, my goal is to make this mixin more generalised, so it could be used for any widget, not just learners. For example, the classic tree viewer widget will eventually have to be merged into one single widget that supports both classification and regression trees. The way I solved this in [OWPythagoreanTree](https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owpythagorastree.py#L360-L385) was with a method that would detect which type of tree was present and then call the appropriate methods to handle the tree type. This meant all the method handlers had to be in the same class when really they are addressing different problems, so that was over 200 LOC for just handlers ([link](https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owpythagorastree.py#L387-L579)).

I believe this approach would generalise well to other widgets and would make code, where there are a lot of handlers, more readable.

pavlin-policar:
I've abandoned this idea for a much simpler approach which is much more understandable. I've also moved this to the main Orange repository since I'll need to modify the base learner widget, and potentially other existing files.

PR #57, opened 2016-12-29
Unique widget - filter unique instances

kernc:
Widget that filters data instances as determined unique on a chosen key.

Blocked by https://github.com/biolab/orange3/pull/1843


lanzagar:
Works for me. Merging into Prototypes. We should consider some tweaks after a bit of use, and where to transfer from Prototypes, etc.
(The label "On instance ties, select:" could maybe be expressed better.)

PR #58, opened 2017-01-04
OWUnique: add auto-commit button

kernc:


lanzagar:
Did you try it?
Drag-and-drop always commits (even when autocommit isn't checked), while changing the dropdown list never updates (even when ac=on).

kernc:
Cheers, mate! :beers:

PR #65, opened 2017-02-13
Add a new 'Data Sets' widget

ales-erjavec:
New prototype widget for fetching data sets from a serverfiles based http index.


orangecontrib/prototypes/widgets/owdatasets.py
```diff
@@ -0,0 +1,514 @@
...
+            log = logging.getLogger(__name__)
+            log.exception("Error while fetching updated index")
+            exc_text = format_exception(er)
+            if not allinfolocal:
+                self.error("Could not fetch data set list: {}"
+                           .format(exc_text))
+            else:
+                self.warning("Could not fetch data sets list, only local "
+                             "cached data sets are shown ({})"
+                             .format(exc_text))
```
lanzagar: I think the message --- _without_ the exception --- should be enough (for now), since this also appears during a "normal" workload (no internet).

orangecontrib/prototypes/widgets/owdatasets.py
```diff
@@ -0,0 +1,514 @@
...
+        allinfolocal = list_local()
+        try:
+            res = f.result()
+        except Exception as er:
+            log = logging.getLogger(__name__)
+            log.exception("Error while fetching updated index")
+            exc_text = format_exception(er)
+            if not allinfolocal:
+                self.error("Could not fetch data set list: {}"
+                           .format(exc_text))
```
lanzagar: I think the message --- _without_ the exception --- should be enough (for now), since this also appears during a "normal" workload (no internet).

ales-erjavec: Ok. Done.

PR #70, opened 2017-03-16
OWCorrelations: Use heuristic to get the most promising attribute pairs

VesnaT:
When dataset is too big (i.e too many attributes or instances), it is impossible to calculate correlations for all attribute pairs. KMeans is used to suggest the most promising attribute pairs by selecting the pair within a cluster.

### DIFF
--- a/orangecontrib/prototypes/widgets/owcorrelations.py
+++ b/orangecontrib/prototypes/widgets/owcorrelations.py
@@ -2,8 +2,7 @@ from enum import IntEnum
 from operator import attrgetter
 
 import numpy as np
-from Orange.widgets.utils.signals import Input, Output
-from scipy.stats import spearmanr
+from scipy.stats import spearmanr, pearsonr
 
 from AnyQt.QtCore import Qt, QItemSelectionModel, QItemSelection, QSize
 from AnyQt.QtGui import QStandardItem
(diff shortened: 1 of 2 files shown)


BlazZupan:
Please sort the rows by absolute value of correlations. Also, keep the same precision when displaying numbers (e.g., instead of 0.76 write 0.760).

BlazZupan:
Please show progress bar.

VesnaT:
Do not merge until #2118 and #2117 are merged.

ajdapretnar:
Since both necessary commits are merged, can we merge this one as well? @BlazZupan

astaric:
@BlazZupan, this has been sitting here for more than a year. Is there still need for this widget? Do we ask @VesnaT to rebase, or do we close it as won't fix?

ajdapretnar:
@VesnaT Could you please rebase so we can merge and hopefully include it in the migration of the widget? Thanks!

BlazZupan:
It would be great if we can have this widget finally in Orange.

VesnaT:
It's rebased, but the tests don't pass (I guess since vizrank has been changed). The tests should be fixed before the widget is moved to Orange.

ajdapretnar:
Yeah, it's probably #155.

VesnaT:
The last commit fixes #155

ajdapretnar:
I have tried this and while it looks really good, I am getting a segfault when switching between Pearson and Spearman. I cannot easily replicate this, but it happens each and every time when using iris.tab, clicking on different feature pairs and switching between Pearson and Spearman. Normally I select a feature pair and then switch the method.

ajdapretnar:
I merged this. I will have a look at the segfault issue and try to have it as replicable as possible. This will be a separate issue.

PR #72, opened 2017-04-10
[ENH] OWFeatureStatistics

pavlin-policar:
Show basic statistics for every feature.

Features of note:
- Widget is usable with large datasets (means, vars and missing are computed on init, hisograms are computed and drawn as needed and cached) 
- Histograms keep their aspect ratio as the widget resizes (however, this ratio is ignored if the height is too small).
- Dropdown to enable coloring by any target variable, or disable coloring entirely.
- Filter textbox for variable names
- Sparse support

Todo:
- Sorting (currently uses slow `QSortFilterProxyModel`), special sorting by variable type, special sorting by first and second moment
- Implement report
- Attribute selection sends dataset with these attributes to output.

Current state (_housing_ and _GDS1615_, respectively):

### DIFF
--- a/orangecontrib/prototypes/widgets/owfeaturestatistics.py
+++ b/orangecontrib/prototypes/widgets/owfeaturestatistics.py
@@ -1,0 +1,604 @@
+"""
+
+TODO:
+  - Sorting by standard deviation: Use coefficient of variation (std/mean)
+    or quartile coefficient of dispersion (Q3 - Q1) / (Q3 + Q1)
+  - Standard deviation for nominal: try out Variation ratio (1 - n_mode/N)
+"""
+import locale
+from enum import IntEnum
+from functools import partial
+from typing import Any, Optional  # pylint: disable=unused-import
+
+import numpy as np
+import scipy.stats as ss
+from AnyQt.QtCore import Qt, QSize, QRectF, QVariant, \
+    QModelIndex, pyqtSlot, QRegExp
+from AnyQt.QtGui import QPainter, QColor, QKeySequence
+from AnyQt.QtWidgets import QStyleOptionViewItem, QShortcut
+from AnyQt.QtWidgets import QStyledItemDelegate, QGraphicsScene, \
+    QTableView, QHeaderView, QStyle
+
+import Orange.statistics.util as ut
+from Orange.canvas.report import plural
+from Orange.data import Table, StringVariable, DiscreteVariable, \
+    ContinuousVariable, TimeVariable, Domain, Variable
+from Orange.widgets import widget, gui
+from Orange.widgets.settings import Setting, ContextSetting, \
+    DomainContextHandler
+from Orange.widgets.utils.itemmodels import DomainModel, AbstractSortTableModel
+from Orange.widgets.utils.signals import Input
+from orangecontrib.prototypes.widgets.utils.histogram import Histogram
+
+
+class FeatureStatisticsTableModel(AbstractSortTableModel):
+    DistributionRole = next(gui.OrangeUserRole)
+
+    CLASS_VAR, META, ATTRIBUTE = range(3)
+    COLOR_FOR_ROLE = {
+        CLASS_VAR: QColor(160, 160, 160),
+        META: QColor(220, 220, 200),
+        ATTRIBUTE: QColor(255, 255, 255),
+    }
+
+    class Columns(IntEnum):
+        ICON, NAME, DISTRIBUTION, CENTER, DISPERSION, MIN, MAX, MISSING = range(8)
+
+        @property
+        def name(self):
+            return {self.ICON: '',
+                    self.NAME: 'Name',
+                    self.DISTRIBUTION: 'Distribution',
+                    self.CENTER: 'Center',
+                    self.DISPERSION: 'Dispersion',
+                    self.MIN: 'Min.',
+                    self.MAX: 'Max.',
+                    self.MISSING: 'Missing values',
+                    }[self.value]
+
+        @property
+        def index(self):
+            return self.value
+
+        @classmethod
+        def from_index(cls, index):
+            return cls(index)
+
+    def __init__(self, data=None, parent=None):
+        """
+
+        Parameters
+        ----------
+        data : Table
+        parent
+        """
+        super().__init__(parent)
+        self._data = data
+        self._domain = domain = data.domain  # type: Domain
+
+        self._attributes = domain.attributes + domain.class_vars + domain.metas
+        self.n_attributes = len(self._attributes)
+        self.n_instances = len(data)
+
+        self.__compute_statistics()
+
+    @staticmethod
+    def _attr_indices(attrs):
+        # type: (List) -> Tuple[List[int], List[int], List[int], List[int]]
+        """Get the indices of different attribute types eg. discrete."""
+        disc_var_idx = [i for i, attr in enumerate(attrs) if isinstance(attr, DiscreteVariable)]
+        cont_var_idx = [i for i, attr in enumerate(attrs)
+                        if isinstance(attr, ContinuousVariable)
+                        and not isinstance(attr, TimeVariable)]
+        time_var_idx = [i for i, attr in enumerate(attrs) if isinstance(attr, TimeVariable)]
+        string_var_idx = [i for i, attr in enumerate(attrs) if isinstance(attr, StringVariable)]
+        return disc_var_idx, cont_var_idx, time_var_idx, string_var_idx
+
+    def __compute_statistics(self):
+        # We will compute statistics over all data at once
+        matrices = [self._data.X, self._data._Y, self._data.metas]
+
+        # Since data matrices can of mixed sparsity, we need to compute
+        # attributes separately for each of them.
+        matrices = zip([
+            self._domain.attributes, self._domain.class_vars, self._domain.metas
+        ], matrices)
+        # Filter out any matrices with size 0, filter the zipped matrices to 
+        # eliminate variables in a single swoop
+        matrices = list(filter(lambda tup: tup[1].size, matrices))
+
+        def _apply_to_types(attrs_x_pair, discrete_f=None, continuous_f=None,
+                            time_f=None, string_f=None, default_val=np.nan):
+            """Apply functions to variable types e.g. discrete_f to discrete 
+            variables. Default value is returned if there is no function 
+            defined for specific variable types."""
+            attrs, x = attrs_x_pair
+            result = np.full(len(attrs), default_val)
+            disc_var_idx, cont_var_idx, time_var_idx, str_var_idx = self._attr_indices(attrs)
+            if discrete_f and x[:, disc_var_idx].size:
+                result[disc_var_idx] = discrete_f(x[:, disc_var_idx].astype(np.float64))
+            if continuous_f and x[:, cont_var_idx].size:
+                result[cont_var_idx] = continuous_f(x[:, cont_var_idx].astype(np.float64))
+            if time_f and x[:, time_var_idx].size:
+                result[time_var_idx] = time_f(x[:, time_var_idx].astype(np.float64))
+            if string_f and x[:, str_var_idx].size:
+                result[str_var_idx] = string_f(x[:, str_var_idx].astype(np.object))
+            return result
+
+        self._variable_types = [type(var) for var in self._attributes]
+        self._variable_names = [var.name.lower() for var in self._attributes]
+
+        # Compute the center
+        _center = partial(
+            _apply_to_types,
+            discrete_f=lambda x: ss.mode(x)[0],
+            continuous_f=lambda x: ut.nanmean(x, axis=0),
+        )
+        self._center = np.hstack(map(_center, matrices))
+
+        # Compute the dispersion
+        def _entropy(x):
+            p = [ut.bincount(row)[0] for row in x.T]
+            p = [pk / np.sum(pk) for pk in p]
+            return np.fromiter((ss.entropy(pk) for pk in p), dtype=np.float64)
+        _dispersion = partial(
+            _apply_to_types,
+            discrete_f=lambda x: _entropy(x),
+            continuous_f=lambda x: ut.nanvar(x, axis=0),
+        )
+        self._dispersion = np.hstack(map(_dispersion, matrices))
+
+        # Compute minimum values
+        _max = partial(
+            _apply_to_types,
+            discrete_f=lambda x: ut.nanmax(x, axis=0),
+            continuous_f=lambda x: ut.nanmax(x, axis=0),
+        )
+        self._max = np.hstack(map(_max, matrices))
+
+        # Compute maximum values
+        _min = partial(
+            _apply_to_types,
+            discrete_f=lambda x: ut.nanmin(x, axis=0),
+            continuous_f=lambda x: ut.nanmin(x, axis=0),
+        )
+        self._min = np.hstack(map(_min, matrices))
+
+        # Compute # of missing values
+        _missing = partial(
+            _apply_to_types,
+            discrete_f=lambda x: ut.countnans(x, axis=0),
+            continuous_f=lambda x: ut.countnans(x, axis=0),
+            string_f=lambda x: (x == StringVariable.Unknown).sum(axis=0),
+            time_f=lambda x: ut.countnans(x, axis=0),
+        )
+        self._missing = np.hstack(map(_missing, matrices))
+
+    def sortColumnData(self, column):
+        if column == self.Columns.ICON:
+            return self._variable_types
+        elif column == self.Columns.NAME:
+            return self._variable_names
+        elif column == self.Columns.DISTRIBUTION:
+            # TODO Implement some form of sorting over the histograms
+            return self._variable_names
+        elif column == self.Columns.CENTER:
+            return self._center
+        elif column == self.Columns.DISPERSION:
+            # Since absolute values of dispersion aren't very helpful when the
+            # variables occupy different ranges, use the coefficient of
+            # variation instead for a more reasonable sorting
+            dispersion = np.array(self._dispersion)
+            _, cont_var_indices, *_ = self._attr_indices(self._attributes)
+            dispersion[cont_var_indices] /= self._center[cont_var_indices]
+            return dispersion
+        elif column == self.Columns.MIN:
+            return self._min
+        elif column == self.Columns.MAX:
+            return self._max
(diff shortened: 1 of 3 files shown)


orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,394 @@
...
+
+    def __len__(self):
+        return len(self.FIELDS)
+
+    @property
+    def _attr_index(self):
+        return self._data.domain.attributes.index(self.attribute)
+
+    def _get_column(self, filter_nan=True):
+        col = self._data.X[:, self._attr_index]
```
kernc: What if the variable were in Y or metas?
Prefer:
```py
col = self._data.get_column_view(self.attribute)[0]
```

orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,394 @@
...
+from PyQt5.QtGui import QColor, QBrush, QPen
+
+from Orange.data import Table, DiscreteVariable, ContinuousVariable
+from Orange.widgets import widget, gui
+from Orange.widgets.settings import Setting
+
+
+def _scale_to_interval(x, start, stop):
+    x_min, x_max = x.min(), x.max()
+    return (stop - start) * (x - x_min) / (x_max - x_min) + start
```
pavlin-policar: Awesome! This is exactly what I needed.

orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,394 @@
...
+        if self.FIELDS[item].field and hasattr(self, self.FIELDS[item].field):
+            result = getattr(self, self.FIELDS[item].field)
+            return result() if callable(result) else result
+
+    def __getattr__(self, item):
+        """Fallback to the attribute object field if it exists."""
+        if hasattr(self.attribute, item):
+            return getattr(self.attribute, item)
+        else:
+            raise AttributeError('Could not find `%s`' % item)
```
pavlin-policar: Yes, that's true, but then in case the error does actually get raised, the AttributeError message will refer to the `self.attribute`, which can be very confusing. This way, it is clear where the error occurs.

orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,256 @@
...
+    # TODO implement a memoize helper function that only remembers attribute
+    def first_moment(self, attribute, data):
+        if attribute in self.__cache['first']:
+            return self.__cache['first'][attribute]
+
+        if attribute.is_discrete:
+            idx = int(stats.mode(data).mode[0])
+            result = data[idx]
+        elif attribute.is_continuous:
+            result = np.mean(data)
```
nikicc: Please use our own implementation of `mean` (`Orange/statistics/util.py`) which can also handle sparse data sets.

orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,256 @@
...
+            result = np.var(data)
+
+        self.__cache['second'][attribute] = result
+        return result
+
+    def valid(self, attribute, data):
+        if attribute in self.__cache['valid']:
+            return self.__cache['valid'][attribute]
+
+        result = int((~np.isnan(data)).sum())
```
nikicc: Please use our own implementation of `countnans` (`Orange/statistics/util.py`) which can also handle sparse data sets.

orangecontrib/prototypes/widgets/utils/histogram.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        if self.target_var and self.target_var.is_discrete:
+            y = one_hot(self.y)
+            bins = np.arange(self.n_bins)[:, np.newaxis]
+            mask = bin_indices == bins
+            distributions = np.zeros((self.n_bins, y.shape[1]))
+            for bin_idx in range(self.n_bins):
+                distributions[bin_idx] = y[mask[bin_idx]].sum(axis=0)
+        else:
+            distributions = np.bincount(bin_indices.astype(np.int64))
```
nikicc: Please use our own implementation of `bincount` (`Orange/statistics/util.py`) which can also handle sparse data sets.

orangecontrib/prototypes/widgets/owattributes.py
```diff
@@ -0,0 +1,255 @@
...
+        hheader = self.view.horizontalHeader()
+        hheader.setStretchLastSection(False)
+        # hheader.setDefaultSectionSize(100)
+
+        vheader = self.view.verticalHeader()
+        # vheader.setDefaultSectionSize(100)
+
+        # TODO Is there a better way to make the table take up all the space?
+        box = gui.vBox(self.mainArea)
+        box.layout().addWidget(self.view)
```
kernc: This will take up all the space?
```py
self.mainArea.layout().addWidget(self.view)  # ?
```

pavlin-policar:
This PR depends on #2458, for the implementation of `digitize`.

nikicc:
@pavlin-policar the widget crashes if I input some data set with textual features (i.e. StringVariables). We should probably just skip textual features.

screen shot 2017-07-10 at 16 40 45

pavlin-policar:
Yes, I'm also quite sure `TimeVariable` would crash the widget as well, but I'm not entirely sure how to handle them. On the other hand, I'm not sure hiding them would be the best idea either, since one would expect all the features to have some sort of summary.

For a `StringVariable` it may make sense to display the most common string (if it exists) and the number of occurences. Drawing a distribution would make no sense, since the number of possible values will probably be way to high and the bars would be next to invisible.

As for `TimeVariable`, I have basically no idea what we could display for that... Perhaps the date range, but that seems like a reach...

I guess I'll just not display them until we can come up with something better.

nikicc:
> For a StringVariable it may make sense to display the most common string (if it exists) and the number of occurences. Drawing a distribution would make no sense, since the number of possible values will probably be way to high and the bars would be next to invisible.

I'm afraid that in many cases this would yield stopwords like `a`, `the` etc.  Perhaps, should we plot a histogram of number of words inside documents?

kernc:
> As for TimeVariable, I have basically no idea what we could display for that...

Obviously was designed by preschool children, but may otherwise work:
```py
tvar = data.domain['time']
time_col = data.get_column_view(tvar)[0]
tvar.rerp_val(nanmin(time_col))
tvar.rerp_val(nanmax(time_col))
tvar.rerp_val(nanmean(time_col))
...
```

ajdapretnar:
As for TimeVariable, I think we should display a line chart of occurrences in time. I know we can't even handle that in a proper plot, but perhaps this is a good opportunity to think about it. Could we create a heuristic that figures out what is the proper granularity of time instances (minute, hour, day, year?) and plot it in a line chart as sum of instances? Normally I'd like a quick glance at when was a particular event the most frequent.

pavlin-policar:
This PR depends on [#2558](https://github.com/biolab/orange3/pull/2558), for the proper implementation of `countnans`.

mstrazar:
Hi guys, any chances this one would be merged soon? This would be very useful for the single cell project.

pavlin-policar:
The widget itself is already usable, however some features are yet to be implemented (see PR description TODO section as well as more informative information for TimeVariable).

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+        SECOND_MOMENT = auto()
+        MISSING = auto()
+
+        @property
+        def name(self):
+            return {self.ICON: '',
+                    self.NAME: 'Name',
+                    self.DISTRIBUTION: 'Distribution',
+                    self.FIRST_MOMENT: 'First moment',
+                    self.SECOND_MOMENT: 'Second moment',
```
kernc: Use friendlier labels "Mean" and "Std".

Also compute "Min", "Max" and "Median" like `pandas.DataFrame.describe()` does.

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+        elif column == self.Columns.NAME:
+            if role == Qt.DisplayRole:
+                output = attribute.name
+        elif column == self.Columns.DISTRIBUTION:
+            if role == self.DistributionRole:
+                if isinstance(attribute, (DiscreteVariable, ContinuousVariable)):
+                    return attribute, self._data
+        elif column == self.Columns.FIRST_MOMENT:
+            if role == Qt.DisplayRole:
+                output = self._first_moment[row]
```
kernc: To format mean timestamp value as mean formatted time, you can add here:
```py
if attribute.is_time:
    output = attribute.repr_val(output)
```

kernc:
> As for TimeVariable, I think we should display a line chart of occurrences in time. ... Normally I'd like a quick glance at when was a particular event the most frequent.

@ajdapretnar How is this much different from a [continuous histogram](https://camo.githubusercontent.com/da8a83f3916fb0c13bfd8a0d613e630c73f39e70/68747470733a2f2f692e6779617a6f2e636f6d2f35386633316261333133613666393034336564353538383063393839643865372e706e67)?

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+        elif column == self.Columns.SECOND_MOMENT:
+            if role == Qt.DisplayRole:
+                output = self._second_moment[row]
+        elif column == self.Columns.MISSING:
+            if role == Qt.DisplayRole:
+                output = '%d / %d (%d%%)' % (
+                    self._missing[row], self._n_instances,
```
kernc: `self._n_instances` is the same for all features? Then better move the number of instances into the control area info box and show here only the absolute count and percent value.

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+        if role == Qt.BackgroundRole:
+            if attribute in self._domain.attributes:
+                return self.COLOR_FOR_ROLE[self.ATTRIBUTE]
+            elif attribute in self._domain.metas:
+                return self.COLOR_FOR_ROLE[self.META]
+            elif attribute in self._domain.class_vars:
+                return self.COLOR_FOR_ROLE[self.CLASS_VAR]
+
+        elif role == Qt.TextAlignmentRole:
+            return Qt.AlignRight | Qt.AlignVCenter
```
kernc:
```py
if column == self.Columns.NAME:
    return Qt.AlignLeft
```

ajdapretnar:
It's smoothed and it handles sums of instances by datetime. Are you gonna plot each event separately? By hour? Day? Month? Year??? There is no sum-by-some-measure-of-time in histograms (unless event occur at _exactly_ the same time).

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+from Orange.data import Table, StringVariable, DiscreteVariable, \
+    ContinuousVariable, TimeVariable
+from Orange.widgets import widget, gui
+from Orange.widgets.settings import Setting, ContextSetting
+from Orange.widgets.utils.itemmodels import DomainModel
+from Orange.widgets.utils.signals import Input
+from orangecontrib.prototypes.widgets.utils.histogram import Histogram
+
+
+class FeatureStatisticsTableModel(QAbstractTableModel):
```
kernc: You may be able to extend `itemmodels.AbstractSortTableModel` for fast numpy sorting, but not (yet) for filtering.

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,512 @@
...
+        ATTRIBUTE: QColor(255, 255, 255),
+    }
+
+    class Columns(IntEnum):
+        ICON = auto()
+        NAME = auto()
+        DISTRIBUTION = auto()
+        FIRST_MOMENT = auto()
+        SECOND_MOMENT = auto()
+        MISSING = auto()
```
kernc: Avoid `enum.auto` as too recent. Use:
```py
ICON, NAME, DISTRIBUTION, MEAN, STD, MIN, MEDIAN, MAX, MISSING = range(9)
```

orangecontrib/prototypes/widgets/owfeaturestatistics.py
pavlin-policar: While I'd love to add these, quartiles would probably also make sense. However, these labels only make sense with ordinal variables and we need reasonable behaviour for nominal and date variables as well, where these values make little or no sense.

For discrete variables, I compute the mode and entropy for the first and second moment (these make some sort of sense?), and having the labels simply as "Mean" and "Std" would be confusing.

I have no good solution for this yet. Splitting them into separate columns would result in empty cells, which would be a complete waste space, that doesn't seem like a viable option.

pavlin-policar:
For the time variable, I agree with @ajdapretnar; a line chart would be far more appropriate than a histogram. As she said in a histogram, we have no way of determining a good granularity for the bins.

However, how would we color the line charts to show the values of the target variable? For a discrete target, we could simply plot multiple lines, each in a different color, but what about for a continuous target? Coloring the area under the curve might be a good solution... If this turns out to work well, it may be worth testing for continuous variables as well.

orangecontrib/prototypes/widgets/owfeaturestatistics.py
kernc: Isn't entropy as Second moment just as confusing?
Having mode and entropy for nominals is great, but the labels will be more intuitive (i.e. fewer StackExchange duplicates) if set to Mean and Std (or Center and Dispersion, or Mean/Mode and Std/Entropy). You can also tutor the user with `OWWidget.UserAdviceMessages` and header item tooltips. But I'll wager far more people understand the concepts mean, mode, std and (possibly) entropy than first and second moment. I'm normally strongly opposed to dumbing stuff down for idiots, but this really is making it complicated for no good reason. CC: @biolab/orange 

Min and Max _do_ make sense for datetime variables. For nominals, you could show the extreme labels from `Variable.values`. This is sensible when nominal `variable.ordered` (e.g. `few < more < many`). The widget is not too wide to add two more columns.

Re quartiles, they are not as essential as they can be (roughly) visually inferred from the histogram and the aforementioned reference points. 😄

pavlin-policar: Yes, I agree, first and second moment is terrible labels. I really like Center and Dispersion.

Yes, min and max make sense for those cases, but in most cases the discrete variable is unordered.

ajdapretnar:
With the recent master on core and this PR I get:
AttributeError: 'OWFeatureStatistics' object has no attribute 'proxy_model'

orangecontrib/prototypes/widgets/owfeaturestatistics.py
```diff
@@ -0,0 +1,595 @@
...
+        # If the user explicitly types different cases, we assume they know
+        # what they are searching for and account for letter case in filter
+        different_case = (
+            any(c.islower() for c in self.filter_string) and
+            any(c.isupper() for c in self.filter_string)
+        )
+        if not different_case:
+            regex.setCaseSensitivity(Qt.CaseInsensitive)
+
+        self.proxy_model.setFilterRegExp(regex)
```
kernc: Looks like a remnant since https://github.com/biolab/orange3-prototypes/pull/72/commits/2e440239cf6b98db1968d746b38f0f3e4386e074. `AbstractTableModel` unfortunately does not do filtering. So you should probably adapt/remove this.

But I think filtering is nice to have. @pavlin-policar What are your thoughts on it? Would you be willing to amend `AbstractSortTableModel` with filtering capabilities, or wrap `FeatureStatisticsTableModel` in a `QSortFilterProxyModel` just for its filtering?

pavlin-policar: My bad, I must have missed it when converting over to the new `AbstractTableModel`.

I think filtering is a very important feature, since the GEO datasets can have thousands of features and looking for features by scrolling can get tedious, very fast. I will look into it.

pavlin-policar:
@kernc, @nikicc This should now be stable enough for another review. I've also squashed all the commits so it's not a complete mess anymore.

kernc:
Histograms are too high? I'd expect them to be the height of a normal table view line, about 2em at most, likened to [sparklines](https://en.wikipedia.org/wiki/Sparkline).
screenshot_2018-01-05_20-54-26

pavlin-policar:
That's actually a feature :smile: The histogram resizes to take up the entire area that the text doesn't need (this could be useful if you find an interesting histogram and want to take a closer look, you resize the window and make it wider, and the histogram grows). If you open and resize the widget, you'll see it work, the default sizes should look all right.

Since it does look quite extreme on your screenshot, perhaps it may be a good idea to set some kind of a limit to how big it can grow?

kernc:
Resizing in width  is fine, but height? The histogram doesn't shrink smaller when the window is shrunk vertically.

pavlin-policar:
That's true, it is prevented from shrinking into nothing. I'll tinker with fixing the height or limiting it. The thing is that you probably want a compact overview of histograms, so you can see many at a time, and want them to be fairly small, but then one may catch your eye and you'd want to inspect it further and make it bigger. Having a fixed height doesn't allow that.

kernc:
Mouse hover could pop over a larger copy of the histogram for closer
inspection? Or better yet, since the binning is as crude, one could instead
select the attribute and pass it out into a Distributions widget for more
precise inspection.

kernc:
@pavlin-policar Could you possibly fix this row height by the end of the weekend so it gets featured in the [next asap release](https://github.com/biolab/orange3-prototypes/issues/109)? _I_ would but don't know how.

pavlin-policar:
Sure thing! The latest commit limits the max height the histogram can grow to. This way the histograms still maintain a reasonable aspect ratio at a couple window sizes. If it turns out that the growth is pointless, it's easy to throw out.

kernc:
I made a change allowing to color by any variable. Is it ok?

pavlin-policar:
I like this change a lot, however, now when I open the widget, there is no option selected for colouring (see screenshot). I think having the colouring default first target variable would make the most sense.

kernc:
Fixed.

pavlin-policar:
Looks good to me :+1:

PR #85, opened 2017-08-10
[ENH] OWDataSets: all links are now in HTML

BlazZupan:
Before the widget assumed that links have URL: link syntax. Uncommon, plus with rendering that displayed the entire URL. This has been changed so that assumed syntax is that of included HTML in the fields of the .info files.


ajdapretnar:
@BlazZupan I think you need to clean up the commit history. :)

PR #90, opened 2017-09-18
OW1ka: EnKlikAnketa widget

kernc:
Likened to Google Sheets widget. Requires bs4.


ajdapretnar:
Please advise: where can I get the proper link in 1ka that will display the survey? Their GUI is a complete mess... 😒

ajdapretnar:
Ok, I managed to test it somewhat.
It is fairly strange having a tooltip with a 'clickable' url, on which you cannot actually click. The tooltip is hidden as soon as I move the cursor. So I had a hard time finding help for setting up a survey link.

Then I followed the instructions and entered the right url. The survey got loaded (info box displayed, output present), but the url instantly defaulted to previous one and the error got displayed again. Does not happen when running --clear-widget-settings.

ajdapretnar:
One more issue. Not sure whether this is a bug on our side or 1ka's, reporting it here anyway.

My survey:
screen shot 2017-09-19 at 10 27 52


Orange spreadsheet:
screen shot 2017-09-19 at 10 28 00

Instead of Izberi = A, it should be Izberi (Prvi) = A, Izberi (Drugi) = B, Izberi (Tretji) = C. And what are those missing features?

kernc:
Hopefully resolved.

ajdapretnar:
Yup, resolved.
I also like the new UserAdviceMessage. I think it is more informative than the old tooltip. 👍 

One final question/issue - Orange decides one of my columns is a string, even though I think it should be, say, categorical. Should we have the same functionality here as we do in the File widget or should we handle this differently? When importing data, you get somehow stuck with the option Orange 'guessed'.

ajdapretnar:
An idea (not necessarily a good one) - should Edit Domain widget become more versatile and enable changes of attribute types?

kernc:
> Edit Domain widget should become more versatile and enable changes of attribute types

My thoughts exactly.

ajdapretnar:
Oh, and should we treat 'neodgovorjeno vprašanje' as nan or what? I am really not sure on this one...

kernc:
Done.

Issue #92, opened 2017-10-09
EnKlik Anketa: fails on None


ajdapretnar:
Widget fails on an empty survey. Should not fail and provide a better-formatted Error.

kernc:
Can you please example URL.

PR #93, opened 2017-10-09
OW1Ka: fix crash if link contains no survey responses

kernc:
Fixes https://github.com/biolab/orange3-prototypes/issues/92

### DIFF
--- a/orangecontrib/prototypes/widgets/ow1ka.py
+++ b/orangecontrib/prototypes/widgets/ow1ka.py
@@ -23,6 +23,10 @@ from Orange.widgets.utils.webview import WebviewWidget, wait
 log = logging.getLogger(__name__)
 
 
+class DataEmptyError(Exception):
+    pass
+
+
 class URLComboBox(QComboBox):
 
     class TitleShowingPopupDelegate(QStyledItemDelegate):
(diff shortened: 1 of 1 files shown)


orangecontrib/prototypes/widgets/ow1ka.py
```diff
@@ -255,7 +268,10 @@ def apply_domain_edit(self):
 
     def table_from_html(self, html):
         soup = BeautifulSoup(html, 'html.parser')
-        html_table = soup.find_all('table')[-1]
+        try:
+            html_table = soup.find_all('table')[-1]
+        except IndexError:
+            raise DataEmptyError
```
astaric: I'd add (). We usually raise Exception instances, not classes.

kernc: I prefer it this way when the exception has no arguments. It's sugar. https://docs.python.org/3/tutorial/errors.html#raising-exceptions Regex `raise \w+$` shows our use of it is quite broad.

astaric: I love the "one obvious way to do it" approach python has :) Leave it as it is then

ajdapretnar:
The submitted survey still doesn't work with this fix. Seems like the issue was not an empty survey, but there must be a None elsewhere... Anyhow, I have results now and I can't see them in Orange.

kernc:
The changeset _fixes_ the crash when the result data is empty, such as in https://www.1ka.si/podatki/141045/63A99D0B/.

It _doesn't fix_ the NoneType crash due to incompatible (unprecedented!) result set format, such as in https://www.1ka.si/podatki/141025/5730371E/ Dafuq is that?!

kernc:
Should not longer crash on Analysis link.

PR #118, opened 2018-02-01
Oracle sql widget

Juriell:
##### Description of changes

A new widget for oracle SQL

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


orangecontrib/prototypes/widgets/oworaclesql.py
```diff
@@ -0,0 +1,136 @@
...
+        if self.savedUsername is not None:
+            self.username = self.savedUsername
+        self.password = ''
+        if self.savedPwd is not None:
+            self.password = self.savedPwd
+        self.database = ''
+        if self.savedDB is not None:
+            self.database = self.savedDB
+        #Control Area layout
+        sip.delete(self.controlArea.layout())
```
astaric: I would just set the new layout and leave deleting of the old one to the gc.

orangecontrib/prototypes/widgets/oworaclesql.py
```diff
@@ -0,0 +1,136 @@
...
+
+        
+class ORACLESQL(OWWidget):
+
+    name = "Oracle SQL"
+    icon = "icons/OracleSQL.svg"
+    want_main_area = True
+    want_message_bar = True
+    #inputs = []
+    outputs = [("Data", Table)]
```
astaric: This is not needed, outputs are declared below with the new-style class.

orangecontrib/prototypes/widgets/oworaclesql.py
```diff
@@ -0,0 +1,136 @@
...
+except:
+    cx_Oracle = None
+    
+
+        
+class ORACLESQL(OWWidget):
+
+    name = "Oracle SQL"
+    icon = "icons/OracleSQL.svg"
+    want_main_area = True
```
astaric: this value is the same as in OWWidget so it can be omitted

orangecontrib/prototypes/widgets/oworaclesql.py
```diff
@@ -0,0 +1,136 @@
...
+    want_message_bar = True
```
astaric: this value is the same as in OWWidget so it can be omitted

orangecontrib/prototypes/widgets/oworaclesql.py
```diff
@@ -0,0 +1,136 @@
...
+    #inputs = []
+    outputs = [("Data", Table)]
+    description = "Select data from oracle databases"
+    settingsHandler = settings.DomainContextHandler()
```
astaric: Remove this line. Widget will work just fine with the ordinary SettingHandler

orangecontrib/prototypes/widgets/oworaclesql.py
Juriell: I'm deleting the original layout because the line  self.controlArea.setLayout(QHBoxLayout()) is not working without it - The SQL box is being placed under the Database box. 
If there is some other way to place the SQL box to the right of the Database box - i can change.

astaric: Have you considered placing the sql box into the main area instead?

astaric:
@Juriell, I have cleaned up the code. If it still works (please check), this can be merged.

Juriell:
@astaric  thanks i will take a look.

kernc:
Thanks!

ajdapretnar:
@Juriell Thank you for contributing! If you have some extra will and determination, I'd appreciate if you could write a short documentation for the widget. Examples can be found [here](https://github.com/biolab/orange3-prototypes/tree/master/doc/widgets).

PR #137, opened 2018-04-16
Feature statistics improvements

### LINKED ISSUE #133: Feature Statistics: sorting bugs
##### Expected behavior
Column sorting should be based on the selected column, say min, max or dispersion.

##### Actual behavior
Sorting does not work when requested for dispersion.

##### Steps to reproduce the behavior
 I have tried it with the following data set:
http://file.biolab.si/files/tmp/sc-aml-sample.pickle

### LINKED ISSUE #134: Feature Statistics: exclude text features and dates
##### Expected behavior
It is probably hard to characterize text features with dispersion, min, and max (not clear what to show here) so I propose to exclude this types of features in the visualization.

##### Actual behavior
Text features are included, reported are NaN's for a min, max, and dispersion.

### LINKED ISSUE #135: Feature Statistics: allow selection of rows (features)
##### Expected behavior
Selection should behave similarly to Rank widget.

##### Actual behavior
Feature selection is not supported.

pavlin-policar:
##### Description of changes
This turned very big, very fast. Sorry. I'll put other improvements in a separate PR, since this is already big as it is.

- Fixes #133 Sorting now works properly. I've also added variable type grouping, so that the dispersion in continuous variables is not mixed with modes of discrete variables, since it makes no sense to compare these. The sort order and column is stored as a ContextSetting
- Fixes #134 We don't show string variables anymore. After a bit of discussion in https://github.com/biolab/orange3/pull/2999, I realized treating time variables as ordinary continuous variables is perfectly reasonable. Center is the mean date, dispersion is the amount of time passed between the min and max date.
- Fixes #135 Outputs like in Rank. The selection is stored as a ContextSetting.
- Refactored most of the code so it's much clearer now, also improved performance by 1x.
- Fix crash on discrete target with missing values.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owfeaturestatistics.py
+++ b/orangecontrib/prototypes/widgets/owfeaturestatistics.py
@@ -144,7 +144,7 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
         _dispersion = partial(
             _apply_to_types,
             discrete_f=lambda x: _entropy(x),
-            continuous_f=lambda x: ut.nanvar(x, axis=0),
+            continuous_f=lambda x: ut.nanvar(x, axis=0) / ut.nanmean(x, axis=0),
         )
         self._dispersion = np.hstack(map(_dispersion, matrices))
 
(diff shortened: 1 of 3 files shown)


lanzagar:
Looks good.

One minor issue I have is with sorting by Center. Numerical variables are trivial, but for categorical I am not sure how it is implemented now - probably sorts by that column in the data matrix (i.e. by indices of values)? Because the values that appear in the table are not sorted... I think it would be better to sort lexicografically by the displayed values (users don't know the internal representation of these categorical values).

lanzagar:
I am merging this so we don't delay all the changes here and in #138.
The sorting issue should be changed in another PR.

PR #143, opened 2018-04-25
Icon redesign

larazupan:


markotoplak:
Thank you. 

Could you matry to make neighbors more obvious byke another short experiment: try coloring points diagonal to the mark lighter? So that you would have just 4 dark points? That could be more relevant as a smaller part of data set was selected.

sn

PR #150, opened 2018-07-17
Widget ExplainPredictions

AndrejaKovacic:
##### Issue

Widget that computes attribute contributions using shapely value, by employing an approximation algorithm. 

##### Description of changes
Added rudimentary version of the proposed widget.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owexplpred.py
+++ b/orangecontrib/prototypes/widgets/owexplpred.py
@@ -1,10 +1,10 @@
-from AnyQt.QtWidgets import QApplication, QFormLayout, QTableView,  QSplitter, QSizePolicy
+from AnyQt.QtWidgets import (
+    QApplication, QFormLayout, QTableView,  QSplitter, QSizePolicy)
 from AnyQt.QtCore import Qt, QThread, pyqtSlot
 from PyQt5.QtGui import QSizePolicy
 import numpy as np
 import scipy.stats as st
 
-import random
 import sys
 import copy
 import logging
@@ -20,8 +20,8 @@ from Orange.widgets.utils.sql import check_sql_input
 from Orange.base import Model
 from Orange.data import DiscreteVariable, ContinuousVariable, Domain, Table
 from Orange.widgets.utils.concurrent import (
-        ThreadExecutor, FutureWatcher, methodinvoke
-        )
+    ThreadExecutor, FutureWatcher, methodinvoke
+)
 
 
 class Task:
@@ -35,28 +35,41 @@ class Task:
         self.future.cancel()
         concurrent.futures.wait([self.future])
 
+
 class ExplainPredictions(object):
     """
-    Class used to explain individual predictions by determining the importance of attribute values. All interactions between atributes are accounted for by calculating Shapely value.
-    Parameters:
+    Class used to explain individual predictions by determining the importance of attribute values.
+    All interactions between atributes are accounted for by calculating Shapely value.
 
-    :param data: table with dataset 
-    :type: Orange data table
-    :param model: model to be used for prediction
-    :type model: Orange model
-    :param pError: p value of error
-    :type pError: float
-    :param error: desired max error of approximation algorithm
-    :type error: float
-    :param batchSize: size of batch used in prediction, bigger batch speeds up the calculations
-    :type batchSize: int
-    :param maxIter: maximum number of iterations
-    :type maxIter: int
-    :param minIter: minimum number of iterations
-    :type minIter:int
+    Parameters
+    ----------
+    data : Orange.data.Table
+        table with dataset
+    model: Orange.base.Model
+        model to be used for prediction
+    error: float
+        desired error 
+    pError : float
+        p value of error
+    batchSize : int
+        size of batch to be used in making predictions, bigger batch size speeds up the calculations and improves estimations of variance
+    maxIter : int
+        maximum number of iterations per attribute
+    minIter : int
+        minimum number of iterations per attiribute
+    seed : int
+        seed for the numpy.random generator, default is 0
+
+    Returns:
+    -------
+    classValue: float
+        either index of predicted class or predicted value
+    table: Orange.data.Table
+        table containing atributes and corresponding contributions, descending
+
     """
 
-    def __init__(self, data, model, pError=0.05, error=0.05, batchSize=100, maxIter=59000, minIter=100):
+    def __init__(self, data, model, pError=0.05, error=0.05, batchSize=100, maxIter=59000, minIter=100, seed=0):
 
         self.model = model
         self.data = data
@@ -65,14 +78,15 @@ class ExplainPredictions(object):
         self.batchSize = batchSize
         self.maxIter = maxIter
         self.minIter = minIter
-        self.atr_names = DiscreteVariable(name = 'attributes', values = [var.name for var in data.domain.attributes])
+        self.atr_names = DiscreteVariable(name='attributes',
+                                          values=[var.name for var in data.domain.attributes])
+        np.random.seed(seed)
 
-
-    def anytime_explain(self, instance, callback = None):
+    def anytime_explain(self, instance, callback=None):
         dataRows, noAtr = self.data.X.shape
         classValue = self.model(instance)[0]
 
-        # placeholders : steps, mean, sum of squared differences, calcuated contribuitons
+        # placeholders
         steps = np.zeros((1, noAtr), dtype=float)
         mu = np.zeros((1, noAtr), dtype=float)
         M2 = np.zeros((1, noAtr), dtype=float)
@@ -83,21 +97,23 @@ class ExplainPredictions(object):
         batchMxSize = self.batchSize * noAtr
         zSq = abs(st.norm.ppf(self.pError/2))**2
 
-        tiled_inst = Table.from_numpy(instance.domain, np.tile(instance.X, (self.batchSize, 1)), np.full((self.batchSize, 1), instance.Y[0]))
+        tiled_inst = Table.from_numpy(instance.domain,
+                                      np.tile(instance.X, (self.batchSize, 1)), np.full((self.batchSize, 1), instance.Y[0]))
         inst1 = copy.deepcopy(tiled_inst)
         inst2 = copy.deepcopy(tiled_inst)
         iterations_reached = np.zeros((1, noAtr))
 
-        while not(all(iterations_reached[0,:] > self.maxIter)):
-            if not(any(iterations_reached[0,:] > self.maxIter)):
-                a = np.random.choice(atr_indices[0], p=(var[0,:]/(np.sum(var[0,:]))))
+        while not(all(iterations_reached[0, :] > self.maxIter)):
+            if not(any(iterations_reached[0, :] > self.maxIter)):
+                a = np.random.choice(atr_indices[0], p=(
+                    var[0, :]/(np.sum(var[0, :]))))
             else:
-                a = np.argmin(iterations_reached[0,:])
+                a = np.argmin(iterations_reached[0, :])
 
             perm = np.random.choice([True, False], batchMxSize, replace=True)
             perm = np.reshape(perm, (self.batchSize, noAtr))
-            rand_data = self.data.X[random.sample(
-                range(dataRows), k=self.batchSize), :]
+            rand_data = self.data.X[np.random.randint(
+                dataRows, size=self.batchSize), :]
             inst1.X = np.copy(tiled_inst.X)
             inst1.X[perm] = rand_data[perm]
             inst2.X = np.copy(inst1.X)
@@ -106,13 +122,13 @@ class ExplainPredictions(object):
             inst2.X[:, a] = rand_data[:, a]
             f1 = self._get_predictions(inst1, classValue)
             f2 = self._get_predictions(inst2, classValue)
-            
+
             diff = np.sum(f1 - f2)
             expl[0, a] += diff
 
             # update variance
             steps[0, a] += self.batchSize
-            iterations_reached[0,a] += self.batchSize
+            iterations_reached[0, a] += self.batchSize
             d = diff - mu[0, a]
             mu[0, a] += d/steps[0, a]
             M2[0, a] += d*(diff - mu[0, a])
@@ -122,16 +138,17 @@ class ExplainPredictions(object):
                 break
 
             # exclude from sampling if necessary
-            neededIter = zSq * var[0, a] / (self.error**2)    
-            if (neededIter <= steps[0, a]) and (steps[0,a] >= self.minIter) or (steps[0, a] > self.maxIter):
-                iterations_reached[0,a] = self.maxIter + 1
-                
-                    
-        expl[0,:] = expl[0,:]/steps[0,:]
+            neededIter = zSq * var[0, a] / (self.error**2)
+            if (neededIter <= steps[0, a]) and (steps[0, a] >= self.minIter) or (steps[0, a] > self.maxIter):
+                iterations_reached[0, a] = self.maxIter + 1
 
-        #creating return array
-        domain = Domain([self.atr_names], [ContinuousVariable('contributions')])
-        table = Table.from_list(domain, np.asarray(self.atr_names.values).reshape(-1, 1))
+        expl[0, :] = expl[0, :]/steps[0, :]
+
+        # creating return array
+        domain = Domain([self.atr_names], [
+                        ContinuousVariable('contributions')])
+        table = Table.from_list(domain, np.asarray(
+            self.atr_names.values).reshape(-1, 1))
         ordered = np.argsort(np.abs(expl[0]))[::-1]
         table.Y = expl.T[ordered]
         table.X = table.X[ordered]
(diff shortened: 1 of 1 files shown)


orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+        while not(all(iterations_reached[0,:] > self.maxIter)):
+            if not(any(iterations_reached[0,:] > self.maxIter)):
+                a = np.random.choice(atr_indices[0], p=(var[0,:]/(np.sum(var[0,:]))))
+            else:
+                a = np.argmin(iterations_reached[0,:])
+
+            perm = np.random.choice([True, False], batchMxSize, replace=True)
+            perm = np.reshape(perm, (self.batchSize, noAtr))
+            rand_data = self.data.X[random.sample(
+                range(dataRows), k=self.batchSize), :]
```
kernc: `np.random.randint(dataRows, size=self.batchSize)`

And then you don't need `random` module?

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+    :param pError: p value of error
+    :type pError: float
+    :param error: desired max error of approximation algorithm
+    :type error: float
+    :param batchSize: size of batch used in prediction, bigger batch speeds up the calculations
+    :type batchSize: int
+    :param maxIter: maximum number of iterations
+    :type maxIter: int
+    :param minIter: minimum number of iterations
+    :type minIter:int
```
kernc: Docstrings are nice, but prefer [Numpydoc](https://numpydoc.readthedocs.io/en/latest/example.html) or [Google-style](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings as they are more readable in source.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+
+class ExplainPredictions(object):
+    """
+    Class used to explain individual predictions by determining the importance of attribute values. All interactions between atributes are accounted for by calculating Shapely value.
+    Parameters:
+
+    :param data: table with dataset 
+    :type: Orange data table
+    :param model: model to be used for prediction
+    :type model: Orange model
```
kernc: Type is `Orange.data.Table`, `Orange.base.Model`, ...

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+    def __init__(self, data, model, pError=0.05, error=0.05, batchSize=100, maxIter=59000, minIter=100):
+
+        self.model = model
+        self.data = data
+        self.pError = pError
+        self.error = error
+        self.batchSize = batchSize
+        self.maxIter = maxIter
+        self.minIter = minIter
+        self.atr_names = DiscreteVariable(name = 'attributes', values = [var.name for var in data.domain.attributes])
```
kernc: Python coding style asserts no space around function parameters and arguments.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+        return classValue, table
+
+    def _get_predictions(self, inst, classValue):
+        if isinstance(self.data.domain.class_vars[0], ContinuousVariable):
+            #regression
+            return self.model(inst)
+        else:
+            #classification
+            predictions =  (self.model(inst) == classValue) * 1
+            #return self.model(inst, Model.ValueProbs)[1][:,int(classValue)]
```
kernc: No checking-in of commented-out code.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+    name = "Explain Predictions"
+    description = "Calculates attribute contributions to final prediction"
+    #icon = "iconImage.png"
+    priority = 200
+    error = settings.Setting(5)
+    p_val = settings.Setting(5)
+
+    class Inputs:
+        data = Input("Data", Table, default = True)
+        model = Input("Model", Model, multiple = False)
```
kernc: Can fix it all in PyCharm with _Edit > Select All_, followed by _Code > Reformat Code_.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+            predictions =  (self.model(inst) == classValue) * 1
+            #return self.model(inst, Model.ValueProbs)[1][:,int(classValue)]
+            return predictions 
+
+
+
+class OWExplainPred(OWWidget):
+
+    name = "Explain Predictions"
+    description = "Calculates attribute contributions to final prediction"
```
kernc: Computes. To **the**.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+            selectionMode=QTableView.NoSelection,
+            focusPolicy=Qt.StrongFocus
+        )
+
+        box = gui.vBox(self.controlArea, "Stopping criteria")
+        form = QFormLayout()
+        box.layout().addLayout(form)
+
+        self.error_spin = gui.spin(box, self, "error", 1, 100, callback = self._update_error_spin, 
+                    controlWidth = 80, keyboardTracking=False)
```
kernc: https://www.python.org/dev/peps/pep-0008/?#indentation

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+    @check_sql_input
+    def set_sample(self,sample):
+        """Set input 'Sample', checks if size is appropriate"""
+        self.toExplain = sample
+        self.explanations = None
+        self.Warning.selection_not_matching.clear()
+        self.Warning.sample_too_big.clear()
+        if sample is not None:
+            if len(sample.X[0]) != len(self.data.X[0]):
+                self.Warning.selection_not_matching()
```
kernc: Is this a sufficient check that the sample is matching? What about data domain?

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,367 @@
...
+        zSq = abs(st.norm.ppf(self.pError/2))**2
+
+        tiled_inst = Table.from_numpy(instance.domain, np.tile(instance.X, (self.batchSize, 1)), np.full((self.batchSize, 1), instance.Y[0]))
+        inst1 = copy.deepcopy(tiled_inst)
+        inst2 = copy.deepcopy(tiled_inst)
+        iterations_reached = np.zeros((1, noAtr))
+
+        while not(all(iterations_reached[0,:] > self.maxIter)):
+            if not(any(iterations_reached[0,:] > self.maxIter)):
+                a = np.random.choice(atr_indices[0], p=(var[0,:]/(np.sum(var[0,:]))))
```
markotoplak: There is no possibility to use a fixed random seed for your code. For computational code, I would want the option to set the random seed (see how sklearn does it). As for the widget, it should always return the same results by default (use a fixed seed).

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,409 @@
...
+        self.data = None
+        self.model = None
+        self.toExplain = None
+        self.explanations = None
+
+        self._task = None
+        self._executor = ThreadExecutor()
+
+        self.dataview = QTableView(verticalScrollBarPolicy=Qt.ScrollBarAlwaysOn,
+                                   horizontalScrollBarPolicy=Qt.ScrollBarAlwaysOn,
```
lanzagar: Not sure why the scrollbars should be always on...
The widget definitely needs to be bigger, and then I expect horizontal scrolling will not be needed.
Vertical scrolling just when the table is big seems fine to me?

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,409 @@
...
+        super().__init__()
+        self.data = None
+        self.model = None
+        self.toExplain = None
+        self.explanations = None
+
+        self._task = None
+        self._executor = ThreadExecutor()
+
+        self.dataview = QTableView(verticalScrollBarPolicy=Qt.ScrollBarAlwaysOn,
```
lanzagar: The table should be sortable by columns.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,409 @@
...
+from AnyQt.QtCore import Qt, QThread, pyqtSlot
+from PyQt5.QtGui import QSizePolicy
+import numpy as np
+import scipy.stats as st
+
+import sys
+import copy
+import logging
+import concurrent.futures
+from functools import partial
```
lanzagar: The first two groups of imports should be swapped.
Python imports first, external libraries second, internal stuff last.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,409 @@
...
+            f2 = self._get_predictions(inst2, classValue)
+
+            diff = np.sum(f1 - f2)
+            expl[0, a] += diff
+
+            # update variance
+            steps[0, a] += self.batchSize
+            iterations_reached[0, a] += self.batchSize
+            d = diff - mu[0, a]
+            mu[0, a] += d/steps[0, a]
```
lanzagar: Try to be consistent in the use of spaces around binary operators. Unless there are multiple operators with different priorities, always using spaces should be fine.
https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

AndrejaKovacic:
Thank you for comments, all of them should now be resolved.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+    class Inputs:
+        data = Input("Data", Table, default=True)
+        model = Input("Model", Model, multiple=False)
+        sample = Input("Sample", Table)
+
+    class Outputs:
+        explanations = Output("Explanations", Table)
+
+    class Warning(OWWidget.Warning):
+        empty_data = widget.Msg("Empty dataset.")
```
lanzagar: Is this never used?

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+        sample_too_big = widget.Msg("Too many samples to explain.")
+        selection_not_matching = widget.Msg(
```
lanzagar: These two should be Errors not Warnings since the widget does nothing in those cases.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+            "Domain of dataset and the sample to be explained does not match")
```
lanzagar: Data and Sample domains do not match.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+        self.model = None
+        self.toExplain = None
```
lanzagar: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

orangecontrib/prototypes/widgets/owexplpred.py
lanzagar: And should be sorted after computing and creating a new one.
Currently I can sort by contribution, but selecting a new sample to be explained creates a new table that is not sorted.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+        super().onDeleteWidget()
+
+
+def domain_equal(domain1, domain2):
+    """checks if two domains have the same attributes and target values"""
+
+    if len(domain1.class_vars) != len(domain2.class_vars):
+        return False
+    for idx in range(len(domain1.class_vars)):
+        if domain1.class_vars[idx].name != domain2.class_vars[idx].name:
```
lanzagar: a) why are you checking classes? isn't a sample without a class an acceptable instance to be explained
b) why are you comparing names? are variable comparisons not appropriate here for some reason?
c) why are you doing pairwise comparisons in a for loop? isn't `domain1.class_vars == domain2.class_vars` enough?
d) if you needed a loop, it would be nicer to iterate over pairs with zip: `for v1, v2 in zip(domain1.class_vars, domain2.class_vars)` try to avoid `range(len(...` when you do not need the actual indices and can use the elements directly instead

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+            assert self._task.future.done()
+            # disconnect the `_task_finished` slot
+            self._task.watcher.done.disconnect(self._task_finished)
+            self._task_finished(self._task.future)
+
+    def _print_prediction(self, classValue):
+        """
+        Parameters
+        ----------
+        classValue: float
```
lanzagar: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

(and all occurances of this variable, and maybe some other ones too...)

orangecontrib/prototypes/widgets/owexplpred.py
AndrejaKovacic: This function is made because testing for equality  with == would mean the variable from one domain were copies of variables from the other. I was trying to account for some weird case where sample to be explained is not just a selected row from dataset that is the input for widget, but comes from another dataset, that still has the same domain. In that case, vars wouldn't be copies and == would return false.
If there won't be such a case, i'll change it to ==.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def __init__(self, data, model, pError=0.05, error=0.05, batchSize=100, maxIter=59000, minIter=100, seed=0):
+        self.model = model
+        self.data = data
+        self.pError = pError
+        self.error = error
+        self.batchSize = batchSize
+        self.maxIter = maxIter
+        self.minIter = minIter
+        self.atr_names = DiscreteVariable(name='attributes',
```
markotoplak: `atr_names` should be a StringVariable.

Another question is if the output table wouldn't be nicer transposed (column would be continuous attributes with names same as input columns).

orangecontrib/prototypes/widgets/owexplpred.py
markotoplak: Well, if you opened the same kind of file in Orange the Variable descriptors would most likely get reused.

As your code assumes that columns have the same meaning in both data and instance just looking at the name is not enough. You would definitely need equality.

I would try the same approach we use for predictions: convert into other domain. Like starting with (I assume that instance is Table with only one row) `instance = instance.transform(data.domain)` or vice versa. The values Orange will not know how to match will be set to unknown. This would also handle the case if data is somehow (pre)processed and instance is raw, unprocessed instance.

AndrejaKovacic: Thank you Marko, seems like an elegant solution, will do that.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,401 @@
...
+        var = np.ones((1, noAtr), dtype=float)
+
+        atr_indices = np.asarray(range(noAtr)).reshape((1, noAtr))
+        batchMxSize = self.batchSize * noAtr
+        zSq = abs(st.norm.ppf(self.pError/2))**2
+        atr_err = np.zeros((1, noAtr), dtype=float)
+        atr_err.fill(np.nan)
+
+        tiled_inst = Table.from_numpy(instance.domain,
+                                      np.tile(instance.X, (self.batchSize, 1)), np.full((self.batchSize, 1), instance.Y[0]))
```
lanzagar: If the domain has metas this will fail since you provide only X and Y.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,482 @@
...
+    Returns:
+    -------
+    class_value: float
+        either index of predicted class or predicted value
+    table: Orange.data.Table
+        table containing atributes and corresponding contributions
+
+    """
+
+    def __init__(self, data, model, p_val=0.05, error=0.05, batch_size=500, max_iter=100000, min_iter=1000, seed=667892):
```
lanzagar: Why is the default `seed=667892` ?
There are 2 standard options for seeds: 0 or 42 :)

AndrejaKovacic:
With restart option and progress bar comes an error in estimating time needed for completion (ETA). To fix it, ProgressBarMixin should be able to account for such cases.

orangecontrib/prototypes/widgets/owexplpred.py
```diff
@@ -0,0 +1,541 @@
...
+                self.results[key] = None
+        else:
+            self.explanations = results[1]
+            self.commit_output()
+            model = TableModel(self.explanations, parent=None)
+            header = self.dataview.horizontalHeader()
+            model.sort(
+            header.sortIndicatorSection(),
+            header.sortIndicatorOrder())
+            self.dataview.setModel(model)
```
lanzagar: couldn't most of this `else` be replaced with `self.update_view(results[1])`

AndrejaKovacic: Yes, all of it actually, thank you.

ajdapretnar:
Sorry, I just started testing this widget and I have some comments. 🙈 
Could you perhaps try `view.horizontalHeader().setResizeMode(QHeaderView.Stretch)` for resizing the width of the mainArea to the size of table? I used it in Geocoding and it worked, but I'm no Qt expert, so it might not work here.
I also think the table should be sorted by score by default. I think there's a parameter for that somewhere. 🤔 
Finally, I propose to use a more informative error when sample too big, for example: "Can explain only one instance at a time." Or something shorter, but this would hint a user to go with only one data instance. I didn't know I only need one from the beginning. :)

PR #158, opened 2018-08-15
Include data directory in the installation

TrietDHA:
##### Issue
When the `orange3-prototypes` is installed, the `data` directory is missing. This makes the **Webcam Widget crash** when a photo is taken because the `haarcascade_frontalface_default.xml` file in the `data` directory is not found.

##### Description of changes
Add one line in the `setup.py` to include the `data` directory.


kernc:
Thanks!

PR #159, opened 2018-08-27
Graph explain predictions

AndrejaKovacic:
##### Issue
Visualization of prediction explanations

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
This still needs some testing and fixing. I just tried it on iris with a Tree model and there are several problems. See screenshot after opening without changing the size (it looks better when expanded):

iris

For a start, you should probably determine the minimum width each of the 3 column sections of the graph needs and ensure that it can not be resized below that.
I am not sure why only the first feature shows the score visually, and why it does not correspond to the numeric value. Also features are not sorted, despite the default Rank by Absolute contribution setting on the left.

In the code, I am not sure why you changed docstrings from """ to ''' - triple double quotes are the preferred ones. Also one-line docstrings should be on one line (including starting&closing quotes).

I see you also renamed the file (owexplpred -> owexplpredgraph). I wanted to suggest renaming as well, but to something readable - we prefer full names instead of abbreviations (so probably owexplainpredictions.py). The 'graph' part of the name is not needed.

AndrejaKovacic:
About the score values: should there also be one column for error margin of the score?  Also, I'm not sure how useful are exact scores, since they tell the user something mostly by how they relate to eachother.

lanzagar:
I would say it is fine if error margin is available only through the output table.
Even if the scores are mostly useful in a relative manner it is still nice to see the numbers (easy to compare). Otherwise I think many people will be tracing the lines down to the scale and once your eyes go up and down more than three times I start questioning the approach.

ajdapretnar:
Just my 2c, since I'm going through the PRs in the repo. This min size of the widget is too big in my opinion. Example for iris:
screen shot 2018-09-18 at 11 23 56

It takes half my screen and I have a huge screen. 😁 
Also, the blue squares should be unified with our other visualizations. Now they stand out quite a lot. What are these blue squares? That is, what do they show?

AndrejaKovacic:
The boxes show confidence intervals of calculated scores. Do you have any suggestions, how should i change it? Red color seems too harsh, blue was chosen since it goes with the color of progress bar. Do you have an issue with the squares themselves or should i just maybe choose a lighter color, soften the edges?

ajdapretnar:
I have just not seen anything similar in Orange so far, so I think it would be nice to emulate a rectangle we already have somewhere. The one that comes to mind is variance in Box Plot. What do you think?

screen shot 2018-09-18 at 12 17 17

AndrejaKovacic:
I changed the color of boxes and line showing the score. I'm not sure about the score line being blue, but it might be that I'm just used to it being black. What do you think?

ajdapretnar:
I like it better this way. Blue is easier to see, especially if intervals are close to 0 and overlap with the thick black line. Other opinions welcome. :)

PR #160, opened 2018-09-10
[FIX] OWStack: Add default aggregator

### LINKED ISSUE #153: Stacking: Add default aggregator
The stacking widget currently requires an Aggregator learner on input and uses it to stack the provided base learners.
Usually it is good enough (or at least a first approach) to use Logistic Regression for classification tasks and Linear/Ridge Regression for regression. It would be good if the Stacking widget used these by default, when no other specific aggregator model is provided.

robertcv:
##### Issue
Fixes #153 
##### Description of changes
Add StackingFitter to choose the correct default aggregator.
##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/doc/widgets/stacking.rst
+++ b/doc/widgets/stacking.rst
@@ -20,7 +20,7 @@ Outputs
         trained model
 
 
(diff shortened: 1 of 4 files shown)


robertcv:
@lanzagar Can you take a look?

PR #176, opened 2019-03-01
Remove Clock

thocevar:
The widget is rarely/never used. It depends on the timeseries add-on and is in most cases not even loaded unless the user has the timeseries add-on installed. We should move it to timeseries if we want to keep it.


ajdapretnar:
[collective sigh of relief]

PR #181, opened 2019-05-10
OWPivot: New widget

VesnaT:
##### Issue

Add pivot widget.
Implements https://github.com/biolab/orange3/issues/2860

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/icons/Pivot.svg
+++ b/orangecontrib/prototypes/widgets/icons/Pivot.svg
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 22.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 48 48" style="enable-background:new 0 0 48 48;" xml:space="preserve">
+<style type="text/css">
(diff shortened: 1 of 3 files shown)


ajdapretnar:
As few observations:
- sometimes labels are not shown entirely (see screenshot)
Screen Shot 2019-05-15 at 11 17 06

- Show is perhaps a bit confusing, even though there might not be a very nice way how to fix it. Probably it should select the first feature from the list by default (because then selecting different aggregation would change the plot). Also aggregations that do not apply for selected variable should be grayed out? I.e. if categorical attribute is selected in Show, then numeric aggregations are disabled. And the other way around.
- is there a discrete way of delineating different aggregations a bit more? It is ok now, but perhaps a horizontal line between each section would make it even more obvious.
- Show: None makes little sense in my opinion. I think the first feature should be selected by default.

VesnaT:
- I think we've agreed with @janezd to leave all the check boxes enabled. I don't remember why. But I can make irrelevant ones disabled, if preferred.
- Show: I also think the first feature should be selected by default :)
- I'm not sure about the lines to separate the aggregations. Let's try and see..

ajdapretnar:
I like the changes a lot. The separating lines included. Perhaps @janezd can give a final comment, but I like this widget as is (even if my suggestion about disabling aggregations that don't apply is not included).

VesnaT:
@janezd suggested to show a warning instead of disabling the checkboxes. And also to put the widget to core Orange straight away.

ajdapretnar:
Ok and absolutely! You probably need some docs, right? 😬

VesnaT:
I do. :see_no_evil: :blush:

PR #187, opened 2019-09-02
Define test suite

markotoplak:
##### Issue
Running `python setup.py test` executed some other `orangecontrib` tests.

##### Description of changes
Added a test suite.


markotoplak:
Oh, some tests fail, but I did not bother to check why.

PR #193, opened 2020-02-27
[ENH] Create Table widget

robertcv:
This is a simple widget to create a Table on Orange canvas.
If there is a table on the input, then it uses its domain.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
What is the plan with this widget? I think it needs an icon and tests and then it can be merged and potentially moved to Orange.

ajdapretnar:
The icon should actually look like Create Class and there should be a different icon for that widget IMHO.

ajdapretnar:
p.s. Did we not decide to move it to Educational?

PrimozGodec:
Hm. Yep create class icon would fit for this widget but I do not have any idea what new icon for Create Class would look like.

You are right. We talked about moving it to educational.

robertcv:
I will add an icon but I won't have time to further work on this widget. I suggest merging this hear, testing it further, and then potentially move it.

PrimozGodec:
I will merge it now and then after we test it a bit we can move it to educational

PR #195, opened 2020-03-19
OWUnique: Fix settings, refactor and modernize

### LINKED ISSUE #188: Unique: settings should be domain-dependent
Unique crashes when fed new data with a new domain. Clear the settings when new domain is incoming.

janezd:
##### Issue

Fixes #188.

This in an ancient widget, but even in the days of old, settings worked not.

Sending callback via `QTimer` also probably didn't work properly - signals were not processed in the right order.

##### Description of changes

- The problem with settings is fixed by using a list of variables (instead tuple of lists, which `DomainContextHandler` does not understand). Settings are not migrated because they didn't work anyway.
- `set_data` now calls unconditional commit.
- Instead of callback from on drop event, commit is triggered by signals that indicate the model was changed. This makes a class derived from `QListView` unnecessary.
- Signals are migrated to new-style, models no longer wrap, widget preview is implemented as it's done now.
- Minor changes in GUI.

There are still no tests, but the refactored widget should be more testable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owunique.py
+++ b/orangecontrib/prototypes/widgets/owunique.py
@@ -1,24 +1,14 @@
-from collections import OrderedDict
 from operator import itemgetter
 
 import numpy as np
 
-from AnyQt.QtCore import Qt, QTimer
-from AnyQt.QtWidgets import QApplication, QListView
+from AnyQt.QtCore import Qt
+from AnyQt.QtWidgets import QListView, QSizePolicy, QComboBox
 
 from Orange.data import Table
 from Orange.widgets import widget, gui, settings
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This works well for the problem described in #188. But I found another bug, that is present across Orange.
Connect File to Select Columns, then to Unique and set the grouping variable to, say, Iris. Then remove iris from the domain with Select Columns. The widget crashes. It is not the only one to do so, the same thing happens with Edit Domain if I remember correctly.

janezd:
I can't replicate it. Can you post the output?

Issue #203, opened 2020-04-14
Possible improvements of explain prediction widget


PrimozGodec:
- [x] Change default zoom setting to something bigger, the minimum size of the graph stays unchanged
- [x] Empty space above the stripe and below the stripe currently expand when zooming in. It would be better to always show just one tick (on the y-axis) above and one bellow the stripe.
- [x] Tooltip at the prediction output number should have the text: `Model prediction: <number>`
- [x] Tooltip at the base value number should have the text: `Base value: <number> \nThe average prediction for selected class`
- [x] Fix saved graph - currently not complete graph is at the output image.
- [ ] School with ctrl/cmd + wheel
- [x] Add target class to the report

PrimozGodec:
Moved to https://github.com/biolab/orange3-explain/issues/4

PR #205, opened 2020-04-14
[ENH] Explanation widgets updates

VesnaT:
##### Issue

Implements #203

##### Description of changes
- change default zoom setting to something bigger
- (partially) Empty space above the stripe and below the stripe currently expand when zooming in. It would be better to always show just one tick (on the y-axis) above and one bellow the stripe.
- set tooltip at the model prediction number
- set tooltip at the base value number
- fix saved graph
- add target class to the report (explain model and explain prediction widgets)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
Everything is good. :) Thank you

Issue #214, opened 2020-08-20
Create table widget: possible improvements


PrimozGodec:
I have merged the widget to the master of Prototypes to sit here for a while and get tested. After while it will be moved to educational. Here we can comment about possible improvements. My suggestions are:

- Currently, widget removes data when the input data is disconnected and when the same table is reconnected the table is still empty. My idea is to remember data via Context setting and refill the table when the same table is on the input.
- Currently, the user is not allowed to create a table with numbers and words in same column. I would allow that. If only numbers are in the column the variable would be Continuous else it would be discrete.

ajdapretnar:
I would also start with widget with 3x3 or 5x5 default table. 1x1 really doesn't make sense.
When moving around the table with a tab, the fields get cleared. I think they shouldn't. They also get cleared on enter. Strange behavior. I would expect to press enter to exit the modify mode and for arrow keys to work.
Dates don't get recognized as a TimeVariable.

PR #215, opened 2020-09-15
Improvement of create table widget

PrimozGodec:
##### Issue

Implements #214

##### Description of changes
- Make widget remembers the data for the domain in the context
- Increase the initial table size
- Allow mixed numbers and string in the same column
- Support for the time variable
- Fix bug: data removed when line edit opened
- Tests
- Code cleanup

#### TODO:
- Documentation

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owcreatetable.py
+++ b/orangecontrib/prototypes/widgets/owcreatetable.py
@@ -1,3 +1,4 @@
+import copy
 from typing import Optional
 
 import numpy as np
@@ -5,12 +6,20 @@ from AnyQt.QtCore import Qt, QAbstractTableModel, QModelIndex, QSize
 from AnyQt.QtWidgets import QTableView, QItemDelegate, QLineEdit, QCompleter
 
 from Orange.data import (
-    Table, Domain, ContinuousVariable, DiscreteVariable, TimeVariable
+    Table,
+    Domain,
+    ContinuousVariable,
+    DiscreteVariable,
+    TimeVariable,
 )
 from Orange.widgets import gui
 from Orange.widgets.settings import (
-    Setting, ContextSetting, PerfectDomainContextHandler
+    Setting,
+    ContextSetting,
+    PerfectDomainContextHandler,
+    ContextHandler,
 )
+from Orange.widgets.utils import vartype
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.widget import OWWidget, Output, Input, Msg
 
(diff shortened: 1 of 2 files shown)


orangecontrib/prototypes/widgets/owcreatetable.py
```diff
@@ -160,21 +220,44 @@ class Outputs:
...
+        transform_err = Msg("Data does not fit to domain")
+
+    settingsHandler = PerfectDomainContextHandler()
 
-    n_rows = Setting(1)
-    n_columns = Setting(1)
+    n_rows = Setting(len(DEFAULT_DATA))
+    n_columns = Setting(len(DEFAULT_DATA[0]))
     auto_commit = Setting(True)
+    # since data is a small (at most 20x20) table we can afford to store it
```
ajdapretnar, line 230: When input domain is larger than 20 variables, that would not be the case. The number of columns is arbitrarily large. The number of rows, however, is fixed to 20. Is that still a limitation is this case?

ajdapretnar:
This works quite well and I'd be happy to merge. One small thing. When I enter custom data, then feed the widget a domain (say iris), then remove the connection, I would expect to have the previous custom data waiting for me. It happens for the data based on a domain, so I see no reason not to have that for custom data, too.

orangecontrib/prototypes/widgets/owcreatetable.py
PrimozGodec: When the input domain is larger we allow more columns. I think it is OK, nobody would have a patience to fill 100 columns and 20 rowns manually.

PrimozGodec:
> This works quite well and I'd be happy to merge. One small thing. When I enter custom data, then feed the widget a domain (say iris), then remove the connection, I would expect to have the previous custom data waiting for me. It happens for the data based on a domain, so I see no reason not to have that for custom data, too.

Yep, this is the only reason why it is still WIP. I will probably implement that tommorrow.

PrimozGodec:
> This works quite well and I'd be happy to merge. One small thing. When I enter custom data, then feed the widget a domain (say iris), then remove the connection, I would expect to have the previous custom data waiting for me. It happens for the data based on a domain, so I see no reason not to have that for custom data, too.

It should work now. :) from my side widget can be merged if everything is OK

ajdapretnar:
I am merging, since it is a prototype. There is one thing that need to be fixed, but I will open a separate issue for this.

PR #216, opened 2020-09-16
Hierarchical Clustering extended with computation node significance

aturanjanin:
The original algorithm for the node significance computation is implemented in R by Suzuki and Shimodira (2006): Pvclust: an R package for assessing the uncertainty in hierarchical clustering. We did reimplementation in Python of this algorithm and incorporate it in Orange. 

The aim of the Orange implementation is the better visualisation of the results of the mentioned algorithm. Because of that, we needed to update already existing objects and procedures of OW Hierarchical Clustering, OW Distances and files related to these two widgets. The only new file is `pvclust.py` and all the remaining files in this PR are modified versions of already existing files in Orange3. This is done in order to easier incorporate this version of Hierarchical Clustering and Distances into Orange3 in the future, i.e. so that the already existing files would be only updated but not entirely rewritten.

This implementation is part of the Master Thesis at the Faculty of Computer and Information Science, University of Ljubljana.


orangecontrib/prototypes/hierarchicalclusteringplus/distance/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .distance import (Distance, DistanceModel,
+                       Euclidean, Manhattan, Cosine, Jaccard,
+                       SpearmanR, SpearmanRAbsolute, PearsonR, PearsonRAbsolute,
+                       Mahalanobis, MahalanobisDistance, Hamming, Bhattacharyya)
```
ajdapretnar, line 4: I think there's no need to have that separate from Orange. You can simply use:
`from Orange.distance import X, Y, Z`

ajdapretnar:
There's no need to copy so many Orange files. Simply import them from Orange.

aturanjanin:
As I stated in the description of the pull request, many already existing files related to OWDistances and OWHierarchicalClustering needed to be modified.

For example, for OWDistances no visual changes have been made, but in order to have all the necessary attributes on the output, the `DistMatrix` object had to be modified and almost all classes and their methods from files `base.py, distmatrix.py` and `distance.py` had to be modified too and then this modified versions are imported and these are part of this pull request. 

Similar is for the files related to OWHierarchicalClustering - `hierarchical.py` and `dendrogram.py` where attributes are added to `Tree` class and functions that are using this object are modified accordingly. These changes are then used in the class `DendrogramWidget` to update the drawing of the dendrogram when the significance of clusters is computed. That is why I also put these files on their own as the part of the pull request.

Sorry if it wasn't made clear enough in my previous description.

PR #218, opened 2020-09-24
Quick fix for LDA

### LINKED ISSUE #191: LDA Widget is not working
##### Prototypes version
0.12.0

##### Orange version

3.24.1
##### Actual behavior

LDA widget is not working, in the past we found this tool pretty useful, for instance to perform analyses like PC-LDA.

##### Steps to reproduce the behavior
We can reproduce the error, for instance using the Iris data set where there is a pretty clear target variable.

##### Additional info (worksheets, data, screenshots, ...)
image

##### Message Error
image

ajdapretnar:
##### Issue

Partially fixes #191.

##### Description of changes
Basic fixes and modernization of the widget.
Did use try except for negative eigenvalues, because as I understand from the internet, nothing can be done about it. We could use a different solver (svd) in this case and issue a warning. However, for titanic even svd is useless (produces a singular component).
In short, the widget doesn't work on titanic and zoo. It also doesn't work on heart-disease, which it probably could if we ignored discrete values. Perhaps continuization is useless here.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
Works on iris, breaks on heart-disease, which has a binary class, with:
`ValueError: n_components cannot be larger than min(n_features, n_classes - 1).`
(the widget always tries to use 2 components).
I suggest we expose the `n_components` parameter in the widget with appropriate limits.

janezd:
This widget was a prototype and today we decided to start a new one, so the current version is being removed.

PR #221, opened 2020-11-05
Remove odl explain predictions widget

PrimozGodec:
##### Issue

There is now a newer explain widget in Prototypes which is based on SHAP library and is faster, this old widget now confuse users.

##### Description of changes
Removing the old widget. If we would like to re-add it we can revert this PR.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
This looks good. Perhaps remove the icon, too? Just to have a lean repo. :)

PrimozGodec:
I didn't remove the icon since it is currently used as a temporary icon for explain models, explain prediction.

ajdapretnar:
Yes, but these have their own files? So ExplainPredictions.svg can be removed, while ExplainModel.svg and ExplainPred.svg stay?

PrimozGodec:
@ajdapretnar didn't notice that. :) I removed the icon.

PR #225, opened 2020-11-16
Icon redesign

larazupan:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
I merged this PR. There is still an issue with the explain prediction icon. When I open it in Orange or in any other software (Chrome, Firefox) there are visible gaps between triangles and rectangle.

Screenshot 2020-11-16 at 15 33 13

I opened the icon in Inkscape and saw that triangles exactly contact rectangles and echo others. Also, they od not have a border (width 0). Even if I overlap the triangle and rectangle gaps remain. @larazupan do you know what would be the reason for that? Is it an issue with SVG format?

PR #226, opened 2020-11-19
Explain prediction icon fix

larazupan:


PrimozGodec:
This one is ok now. Thank you for your fast response.

PR #232, opened 2021-02-05
Fixes for migration to github actions

janezd:
##### Issue

- Test discovery loaded a directory that did not exist
- All tests were executed twice due to duplication in test discovery
- Create Table called deprecated method `Domain.__bool__`
- Grep used an incorrect file path.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,12 @@
+[run]
+source = orangecontrib/prototypes
+omit =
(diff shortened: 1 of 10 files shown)


janezd:
This is ready, but has to wait until we fix the problem with resolution on Windows (https://github.com/biolab/orange3/issues/5235).

PR #233, opened 2021-02-17
Hub: remove widget

janezd:
##### Issue

Hub widget was mostly a joke (https://orangedatamining.com/blog/2015/09/11/hubbing-with-hub-widget/), and its implementation was a bit of a hack, too.

To extent to which it was useful - switching between data sources for multiple widgets - the Table widget will adequately replace it after https://github.com/biolab/orange3/pull/5233 is merged.

##### Includes
- [X] Code changes


janezd:
We won't remove it. Perhaps just move it to the attic.

PR #237, opened 2021-02-19
Remove LDA and Significant Groups

janezd:
At today's discussion, we decided to remove these two widgets as not being useful currently and unlikely to be nurtured into usefulness ever.


orangecontrib/prototypes/widgets/owsignificantgroups.py
```diff
@@ -1,340 +0,0 @@
...
-    ThreadExecutor, FutureWatcher, methodinvoke
-)
-
-from orangecontrib.prototypes.significance import (
-    perm_test, hyper_test, chi2_test, t_test,
-    fligner_killeen_test, mannwhitneyu_test,
-    gumbel_min_test, gumbel_max_test,
-    CORRECTED_LABEL,
-)
-from orangecontrib.prototypes.pandas_util import table_from_frame
```
kernc, line 30: Missed these two modules, now unused.

janezd: And the icon, too. Thanks!

PR #241, opened 2021-02-22
Unique: Fix crash on domain change; use auto summaries

janezd:
##### Issue

Widget silently crashes if domain is changed when the list view has a selection. Same as https://github.com/biolab/orange3/pull/5354

##### Description of changes

Remove selection before changing the model, not afterwards.

If you select something, change the data and change it back, the selection that you made on the first data is properly retrieved, but the list view is not properly updated and looks as if nothing was selected. This is a general issue, and is fixed in https://github.com/biolab/orange-widget-base/pull/147.

##### Includes
- [X] Code changes

### DIFF
--- a/orangecontrib/prototypes/widgets/owunique.py
+++ b/orangecontrib/prototypes/widgets/owunique.py
@@ -7,7 +7,6 @@ from AnyQt.QtCore import Qt
 from Orange.data import Table
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils.itemmodels import DomainModel
-from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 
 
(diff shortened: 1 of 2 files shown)


ajdapretnar:
There's an issue with this PR. When changing the Groupby attribute, the output is not updated. Only when selecting "Instance to select in each group", the table is updated.
To reproduce:
Iris - Unique - Data Table. Observe DT. Change Unique to grouping just by iris. Nothing changes in DT.

janezd:
> There's an issue with this PR. When changing the Groupby attribute, the output is not updated. Only when selecting "Instance to select in each group", the table is updated.

Fixed by removing the last commit (in which I, kudos to me, overlooked `lambda` and "fixed" the callback by changing `lambda: self.commit()` to `lambda. self.commit`)

ajdapretnar:
This seems to still fail.
Use iris, Unique, then change the data set. This should be a key test for this widget, IMO.

orangecontrib/prototypes/widgets/owunique.py
```diff
@@ -61,30 +60,21 @@ def __init__(self):
     def set_data(self, data):
         self.closeContext()
         self.data = data
+        self.selected_vars = []
```
janezd, line 63: This line is the actual fix. Everything else is about removing summaries. I accidentaly squashed the two commits and I'm too lazy to separate them (because removed lines 71 and 72 belong to different commits :).

PR #242, opened 2021-03-20
Shopping List: Prepare for graduation from prototypes

janezd:
**TL;DR:** I basically rewrote the widget -- not because anybody needed it in the last five years, but because it was here and we though there wouldn't be much to do finish it. :)

**The last commit removes the widget,** because the widget is moved to core widgets. Don't squash before merging, please. :)

- Widget was written five years ago and doesn't use newer signals, errors etc.
- Computation was inefficient, at least in time; now it's better, but works only for dense data (so did the previous one, too).
- I added option to use only numeric variables and to exclude zeros.
- The widget used to show all variables and complained if the user selected one that can't be used; not it shows only the useful ones.
- The function that made the actual computation is now refactored into smaller methods.
- I added tests.

Icon also already exists, so we only need documentation.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
We need to decide whether this goes into Associate add-on or can such transformation be useful generally. What it does, basically is turn Zoo into this:

Screenshot 2021-03-20 at 15 11 50

The buffalo bought hair and teeth and two legs, and so did the calf. There are of course other options, this particular output is achieved by:

Screenshot 2021-03-20 at 15 14 41

Plus, by selecting a subset of animals because names are not unique and the widget requires unique ids.

While it may seem that this belongs to Associate, I suspect that we would often need a widget that does exactly the opposite. And if the opposite widget were among core widgets, this would belong there, too.

irgolic:
So this widget is basically dense to COO sparse conversion?

janezd:
> So this widget is basically dense to COO sparse conversion?

Yes.

After the latest push, it's also sparse to coo sparse conversion. :) I was too lazy to  support for sparse data, but then I realized that `tocoo` already does most of the job. I just needed to add some fancy indexing to remove and re-numerate some columns.

kernc:
This is called [wide and narrow data](https://en.wikipedia.org/wiki/Wide_and_narrow_data). Conversion between the two is really useful in preprocessing.

janezd:
So @irgolic will now write that we should rename this widget to Melt. I'll agree. Which is a shame, because we already have an  icon and I like it.

irgolic:
I've not heard of Melt before, but we should make an effort toward not reinventing the wheel when it comes to data transformation. Wherever possible, we should adopt commonly used terms from the Dataframe interface; it's the most widespread in the ecosystem. That'll shut up anyone who says Orange doesn't teach transferrable skills :)

PR #245, opened 2022-01-31
Quick Select widget

janezd:
##### Issue

This widget would be useful in many cases in which Select Columns is an overkill and also less practical.

Screenshot 2022-01-31 at 17 47 32

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
Three small comments:
- The widget opens very tall initially. Perhaps a sizeHint or make it non-resizable?
- Matching Data could be set as a default output. It would simplify connecting to other widgets.
- Values in Value combobox are not reset when disconnecting the signal. Vesna told me, it is not neat to have it this way. 😬 This is also a problem if I rename the value that is in the comboBox in Edit Domain. Then the widget crashes.

janezd:
- I fixed resizing. Thanks for the hint.
- Matching data of course has to be default. Done.
- Combos were supposed to be cleaned; leaving them would be sloppy. There were bigger problems with contexts, though. (Details are interesting: the value from the second combo cannot be a context setting because the combo is filled only after setting the value in the first combo and this prohibits `openContext` to set both at once. The second combo is copied to a temporary `pending_value`.)
- I also added reporting. Every well-behaved widget should have it.

Tests and documentation are still TBD, but this -- if it works -- should be acceptable for prototypes.

orangecontrib/prototypes/widgets/owquickselect.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def send_report(self):
+        if not self.data:
+            return
+
+        self.report_items(
+            "",
+            [("Filter", f"{self.variable.name} = '{self.value}'"),
+             ("Matching instances",
+              f"{self.n_matched} (out of {len(self.data) - self.n_matched})")]
```
ajdapretnar, line 134: This is, in my opinion, incorrect.

It should say `f"{self.n_matched} (out of {len(self.data)})"`.

janezd: "In my opinion" was entirely unnecessary. "Obviously" would be quite in place and well deserved.

I guess I initially planned to show the number of matching and unmatched.

PR #246, opened 2022-03-02
[ENH] Interactions widget

needNewUsername:
##### Description of changes
Adds widget for calculation of attribute interactions based on Orange's correlations widget. The returned results are consistent with the Interaction Graph widget from Orange2, however this is somewhat dependant on the discretization of the input data (this one handles continuous data via equal-frequency discretization). 

To simplify the code this widget inherits directly from owcorrelations, so the following Warning is raised:
`RuntimeWarning: subclassing of widget classes is deprecated and will be disabled in the future.
Extract code from OWCorrelations or explicitly open it by adding 'openclass=True' to class definition.
  class OWInteractions(OWCorrelations):`
The solution is probably to construct a combined parent class for both owcorrelations and owinteractions.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owinteractions.py
+++ b/orangecontrib/prototypes/widgets/owinteractions.py
@@ -62,16 +62,16 @@ class Interaction:
 
 class Heuristic:
 	def __init__(self, weights):
-		self.n = len(weights)
-		self.w = weights.copy()
-		self.a = np.arange(self.n)
-		self.a = self.a[np.argsort(self.w)]
+		self.weights = weights
+		self.n_attributes = len(self.weights)
+		self.attributes = np.arange(self.n_attributes)
+		self.attributes = self.attributes[np.argsort(self.weights)]
 
 	def generate_states(self):
 		# prioritize two mid ranked attributes over highest first
-		for s in range(1, self.n * (self.n - 1) // 2):
-			for i in range(max(s - self.n + 1, 0), (s + 1) // 2):
-				yield self.a[i], self.a[s - i]
+		for s in range(1, self.n_attributes * (self.n_attributes - 1) // 2):
+			for i in range(max(s - self.n_attributes + 1, 0), (s + 1) // 2):
+				yield self.attributes[i], self.attributes[s - i]
 
 	def get_states(self, initial_state):
 		states = self.generate_states()
@@ -119,8 +119,8 @@ class InteractionRank(Orange.widgets.data.owcorrelations.CorrelationRank):
 		for i, attr in enumerate(attrs):
 			item = QStandardItem(attr.name)
 			item.setData(attrs, self._AttrRole)
-			item.setData(Qt.AlignLeft + Qt.AlignTop, Qt.TextAlignmentRole)
-			item.setToolTip("Attribute Info: {:.1f}%".format(100*score[2+i]))
+			item.setData(Qt.AlignLeft + Qt.AlignCenter, Qt.TextAlignmentRole)
+			item.setToolTip("{}\nInfo Gain: {:.1f}%".format(attr.name, 100*score[2+i]))
 			attr_items.append(item)
 		interaction_item = QStandardItem("{:+.1f}%".format(100*score[1]))
 		interaction_item.setData(score[1], self.IntRole)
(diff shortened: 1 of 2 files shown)


VesnaT:
Since the code is mostly copied from correlations, the latest correlations fix could also be applied:
https://github.com/biolab/orange3/pull/5869

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -0,0 +1,242 @@
...
+		button_box = gui.hBox(self.buttonsArea)
+		button_box.layout().addWidget(self.vizrank.button)
+
+	@Inputs.data
+	def set_data(self, data):
+		self.closeContext()
+		self.clear_messages()
+		self.data = data
+		self.disc_data = None
+		self.selection = []
```
VesnaT, line 188: The `selection` setting is also missing.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -0,0 +1,242 @@
...
+			self.NEGATIVE_COLOR if score[1] < 0 else self.POSITIVE_COLOR,
+			gui.TableBarItem.BarColorRole)
+		interaction_item.setToolTip("Entropy removed: {:.1f}%".format(100*score[4]))
+		return [interaction_item] + attr_items
+
+	def check_preconditions(self):
+		return self.master.disc_data is not None
+
+
+class OWInteractions(Orange.widgets.data.owcorrelations.OWCorrelations):
```
VesnaT, line 139: If you extend a widget, you should set its property `openclass=True`.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -0,0 +1,242 @@
...
+		attrs = np.c_[self.data.X[:, attr1], self.data.X[:, attr2]]
+		self.removed_h[attr1, attr2] = self.entropy(attrs) + self.class_h - self.entropy(np.c_[attrs, self.data.Y])
+		return self.removed_h[attr1, attr2] - self.gains[attr1] - self.gains[attr2]
+
+
+class Heuristic:
+	def __init__(self, weights):
+		self.n = len(weights)
+		self.w = weights.copy()
+		self.a = np.arange(self.n)
```
VesnaT, line 67: I'd prefer more descriptive variable names.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -0,0 +1,242 @@
...
+from Orange.widgets.widget import OWWidget, AttributeList, Msg
+from Orange.preprocess import Discretize, Remove
+from Orange.preprocess.discretize import EqualFreq
+import Orange.widgets.data.owcorrelations
+
+
+SIZE_LIMIT = 1000000
+
+
+class Interaction:
```
markotoplak, line 26: The class `Interaction` should be tested separately, independently of widget tests, in its own TestSomething class. The same for heuristics.

markotoplak: What this tests should test is what your results should be. They should also test behavior in strange cases, like unknown values (if the class supports them).

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -0,0 +1,242 @@
...
+		self.data = None  # type: Table
+		self.disc_data = None  # type: Table
+
+		# GUI
+		box = gui.vBox(self.controlArea)
+		self.feature_model = DomainModel(
+			order=DomainModel.ATTRIBUTES, separators=False,
+			placeholder="(All combinations)")
+		gui.comboBox(
+			box, self, "feature", callback=self._feature_combo_changed,
```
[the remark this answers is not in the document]
needNewUsername: the default value has already been set by the parent

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: As far as I know, this is something that needs to changed in the base orange3 repository, not this one. Should I open a pull request there as well?

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: Fixed in the latest commit

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: I've moved the tests to their own classes and added tests for sparse data.

orangecontrib/prototypes/widgets/owinteractions.py
markotoplak: I think we should wait with this one: let's not open the class. Instead, when this widget gets out of the prototype phase, perhaps create a common superclass for both.

PR #249, opened 2022-03-21
Interaction visualisation

needNewUsername:
Created new table bar item delegate to visualize interactions and added option to select heuristic.


orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -113,28 +219,41 @@ def compute_score(self, state):
...
-		interaction_item.setData(
-			self.NEGATIVE_COLOR if score[1] < 0 else self.POSITIVE_COLOR,
-			gui.TableBarItem.BarColorRole)
-		interaction_item.setToolTip("Entropy removed: {:.1f}%".format(100*score[4]))
 		return [interaction_item] + attr_items
 
 	def check_preconditions(self):
 		return self.master.disc_data is not None
 
+	def stopped(self):
```
markotoplak, line 251: Does this really fit in stopped? It seems like these things should be set even before stopping.

markotoplak: Ok, I see that `OWCorrelations` has the same pattern, so feel free to go forward with the same. But still, do not reimplement the whole function. Instead, call `super().stopped()` and add only the changes.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -113,28 +219,41 @@ def compute_score(self, state):
 		removed = self.interaction.removed_h[attr1, attr2] / h
 		return -score, score, gain1, gain2, removed
 
+	def _update_model(self):
```
markotoplak, line 222: I'd avoid the reimplementation here. If your visualizer does not used `gui.TableBarItem.BarRole`, no change is needed.

If you use the role (which I think you should not), then you can stop it being reset by just reimplementing `bar_length()` to return None.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -82,27 +107,108 @@ def get_states(self, initial_state):
...
+		style = QApplication.style() if widget is None else widget.style()
+		self.__style = style
+		text = opt.text
+		opt.text = ""
+		style.drawControl(QStyle.CE_ItemViewItem, opt, painter, widget)
+		textrect = style.subElementRect(
+			QStyle.SE_ItemViewItemText, opt, widget)
+
+		center = self.cachedData(index, InteractionRank.IntRole)
+		if center is not None:
```
markotoplak, line 135: Isn't in this case `center` always defined? If so, why the condition?

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -82,27 +107,108 @@ def get_states(self, initial_state):
...
+			painter.drawLine(line)
+			pen.setColor(color)
+			painter.setPen(pen)
+			line.setLine(
+				rect.left() + 2*hmargin + width * left,
+				baseline,
+				rect.left() + 2*hmargin + width * (left + center),
+				baseline
+			)
+			painter.drawLine(line)
```
markotoplak, line 177: These setline/drawline code pattern repeats a lo. They should be probably converted to an internal function which will be then called three times.

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: Yes it bugged me also, but I couldn't figure out why it doesn't work if I put it in `__init__` or `initialize`. As far as calling `super().stopped()` it really only saves me 1 line of code so I just skipped it. If that's the accepted practice I will rewrite it though.

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: center is only defined for the leftmost column. for the other two columns it is `None` so that the drawing is skipped and the text still appears.

orangecontrib/prototypes/widgets/owinteractions.py
needNewUsername: You are right, I think i changed it before I settled on a final solution and now I no longer need it, so I will remove it completely.

What should I use to get the scores to the delegat instead of Roles?

markotoplak: Roles are fine, but if you want to avoid the effects of that setData statement (that is in VizrankAttrPair but not in your code), just don't use `gui.TableBarItem.BarRole`. Any other Role would work.

orangecontrib/prototypes/widgets/owinteractions.py
markotoplak: Makes sense. That would make a great comment in code. Anytime you use assumptions that are likely unclear to someone reading the code, please make a comment.

orangecontrib/prototypes/widgets/owinteractions.py
```diff
@@ -116,6 +116,11 @@ def __init__(self, **kwargs):
 		self.__line = QLineF()
 		self.__pen = QPen(self.b, 5, Qt.SolidLine, Qt.RoundCap)
 
+	def draw_line(self, painter, baseline, left, right):
```
markotoplak, line 119: This should be a function defined within `paint`, as it is only used there.

orangecontrib/prototypes/widgets/owinteractions.py
markotoplak: I think you explained it much better on Github than in the code: here you also provided some context why would only interaction items draw the bars, there that is left as a question to the poor programmer maintaining this code after a few years... I would appreciate if you expand this comment, but will not block the merge because of it.

markotoplak:
Thanks for the fixes. For additional clarity (of git history), this PR could be rebased into a single commit. :)

PR #252, opened 2022-05-12
OWSplit: new widget

ajdapretnar:
##### Issue

Survey and archaeological data often come with many categorical values separated by a delimiter. Orange had no option to create "one hot encoded" variables by splitting by a delimiter.

##### Description of changes
A widget that takes a delimiter and creates new variables by splitting on it.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owsplit.py
+++ b/orangecontrib/prototypes/widgets/owsplit.py
@@ -20,8 +20,15 @@ class SplitColumn:
 
         column = self.get_string_values(data, self.attr)
         values = [s.split(self.delimiter) for s in column]
-        self.new_values = sorted({val if val else "?" for vals in values for
-                                  val in vals})
+        self.new_values = tuple(sorted({val if val else "?" for vals in
+                                        values for val in vals}))
+
+    def __eq__(self, other):
+        return self.attr == other.attr and self.delimiter == \
+               other.delimiter and self.new_values == other.new_values
+
+    def __hash__(self):
+        return hash((self.attr, self.delimiter, self.new_values))
 
     def __call__(self, data):
         column = self.get_string_values(data, self.attr)
@@ -44,6 +51,13 @@ class OneHotStrings(SharedComputeValue):
     def __init__(self, fn, new_feature):
         super().__init__(fn)
         self.new_feature = new_feature
+
+    def __eq__(self, other):
+        return self.compute_shared == other.compute_shared \
+           and self.new_feature == other.new_feature
+
+    def __hash__(self):
+        return hash((self.compute_shared, self.new_feature))
 
     def compute(self, data, shared_data):
         indices = shared_data[self.new_feature]
--- a/orangecontrib/prototypes/widgets/owsplit.py
+++ b/orangecontrib/prototypes/widgets/owsplit.py
@@ -1,0 +1,135 @@
+import numpy as np
+
+from AnyQt.QtCore import Qt
+
+from Orange.widgets import gui
+from Orange.widgets.settings import ContextSetting, DomainContextHandler
+from Orange.widgets.widget import OWWidget, Msg, Output, Input
+from Orange.widgets.utils.itemmodels import DomainModel
+from Orange.widgets.utils.widgetpreview import WidgetPreview
+from Orange.data import Table, Domain, DiscreteVariable, StringVariable
+from Orange.data.util import SharedComputeValue, get_unique_names
+
+from orangewidget.settings import Setting
+
+
+class SplitColumn:
+    def __init__(self, data, attr, delimiter):
+        self.attr = attr
+        self.delimiter = delimiter
+
+        column = self.get_string_values(data, self.attr)
+        values = [s.split(self.delimiter) for s in column]
+        self.new_values = sorted({val if val else "?" for vals in values for
+                                  val in vals})
+
+    def __call__(self, data):
+        column = self.get_string_values(data, self.attr)
+        values = [set(s.split(self.delimiter)) for s in column]
+        shared_data = {v: [i for i, xs in enumerate(values) if v in xs] for v
+                       in self.new_values}
+        return shared_data
+
+    @staticmethod
+    def get_string_values(data, var):
+        # turn discrete to string variable
+        column = data.get_column_view(var)[0]
+        if var.is_discrete:
+            return [var.str_val(x) for x in column]
+        return column
+
+
+class OneHotStrings(SharedComputeValue):
+
+    def __init__(self, fn, new_feature):
+        super().__init__(fn)
+        self.new_feature = new_feature
+
+    def compute(self, data, shared_data):
+        indices = shared_data[self.new_feature]
+        col = np.zeros(len(data))
+        col[indices] = 1
+        return col
+
+
+class OWSplit(OWWidget):
+    name = "Split"
+    description = "Split string variables to create discrete."
+    icon = "icons/Split.svg"
+    priority = 700
+
+    class Inputs:
+        data = Input("Data", Table)
+
+    class Outputs:
+        data = Output("Data", Table)
+
+    class Warning(OWWidget.Warning):
+        no_disc = Msg("Data contains only numeric variables.")
+
+    want_main_area = False
+    resizing_enabled = False
+
+    settingsHandler = DomainContextHandler()
+    attribute = ContextSetting(None)
+    delimiter = ContextSetting(";")
+    auto_apply = Setting(True)
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        variable_select_box = gui.vBox(self.controlArea, "Variable")
+
+        gui.comboBox(variable_select_box, self, "attribute",
+                     orientation=Qt.Horizontal, searchable=True,
+                     callback=self.apply.deferred,
+                     model=DomainModel(valid_types=(StringVariable,
+                                                    DiscreteVariable)))
+        gui.lineEdit(
+            variable_select_box, self, "delimiter",
+            orientation=Qt.Horizontal, callback=self.apply.deferred)
+
+        gui.auto_apply(self.buttonsArea, self, commit=self.apply)
+
+    @Inputs.data
+    def set_data(self, data):
+        self.closeContext()
+        self.data = data
+
+        model = self.controls.attribute.model()
+        model.set_domain(data.domain if data is not None else None)
+        self.Warning.no_disc(shown=data is not None and not model)
+        if not model:
+            self.attribute = None
+            self.data = None
+            return
+        self.attribute = model[0]
+        self.openContext(data)
+        self.apply.now()
+
+    @gui.deferred
+    def apply(self):
+        if self.attribute is None:
+            self.Outputs.data.send(None)
+            return
+        var = self.data.domain[self.attribute]
+
+        sc = SplitColumn(self.data, var, self.delimiter)
+
+        new_columns = tuple(DiscreteVariable(
+                get_unique_names(self.data.domain, v), values=("0", "1"),
+                compute_value=OneHotStrings(sc, v)
+            ) for v in sc.new_values)
+
+        new_domain = Domain(
+            self.data.domain.attributes + new_columns,
+            self.data.domain.class_vars, self.data.domain.metas
+        )
+        extended_data = self.data.transform(new_domain)
+        self.Outputs.data.send(extended_data)
+
+
+if __name__ == "__main__":  # pragma: no cover
+    WidgetPreview(OWSplit).run(Table.from_file(
+        "tests/orange-in-education.tab"))
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
 recursive-include doc *.bat *.css *.md *.png *.py *.rst Makefile
-recursive-include orangecontrib *.svg icons/*  *.html *.js *.css *.txt
+recursive-include orangecontrib *.svg icons/*  *.html *.js *.css *.txt *.tab
 include *.pypi *.md *.ini .coveragerc
 
--- a/orangecontrib/prototypes/widgets/icons/Split.svg
+++ b/orangecontrib/prototypes/widgets/icons/Split.svg
@@ -1,0 +1,184 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Generator: Adobe Illustrator 15.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+
+<svg
+   version="1.1"
+   id="Layer_1"
+   x="0px"
+   y="0px"
+   width="48px"
+   height="48px"
+   viewBox="0 0 48 48"
+   enable-background="new 0 0 48 48"
+   xml:space="preserve"
+   sodipodi:docname="Split.svg"
+   inkscape:version="1.1.2 (b8e25be8, 2022-02-05)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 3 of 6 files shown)


orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+from Orange.widgets import gui
+from Orange.widgets.settings import ContextSetting, DomainContextHandler
+from Orange.widgets.widget import OWWidget, Msg, Output, Input
+from AnyQt.QtCore import Qt
+
+from Orange.data import DiscreteVariable, StringVariable
+from Orange.widgets.utils.itemmodels import DomainModel
+from Orange.widgets.utils.widgetpreview import WidgetPreview
+
+foo = ["foo,bar", "foo,baz", "foo", "bar,baz", "baz,foo", "baz", "bar"]
```
janezd, line 14: Foo.

janezd, line 14: Foo. Bar, too.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+        X[i, list(map(mapper.get, vals))] = 1
+    return X, new_values
+
+
+def values_to_string(var_list, values):
+    # turn discrete to string variable
+    return [values[int(i)] if not np.isnan(i) else "nan" for i in var_list]
+
+
+class TransformDomain:
```
janezd, line 33: As far as I see, this class is not used. Neither necessary?

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+    attribute = ContextSetting(None)
+    splitter = ContextSetting(";")
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        gui.lineEdit(
+            self.controlArea, self, "splitter",
+            orientation=Qt.Horizontal, box="Split by", callback=self.apply)
```
janezd, line 83: Add auto comit, use deferred commit, and here set `callback` to `self.apply.deferred`.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+
+        variable_select_box = gui.vBox(self.controlArea, "Variable")
+
+        gui.comboBox(variable_select_box, self, "attribute", label=None,
+                     orientation=Qt.Horizontal, searchable=True,
+                     callback=self.apply,
```
janezd, line 89: `self.apply.deferred`

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+    @Inputs.data
+    def set_data(self, data):
+        self.closeContext()
+        self.data = data
+
+        model = self.controls.attribute.model()
+        model.set_domain(data.domain if data is not None else None)
+        self.Warning.no_disc(shown=data is not None and not model)
+        if not model:
+            self.attribute = None
```
janezd, line 102: Also set `self.data` to `None` to ensure the widget does not do anything.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+            self.Outputs.data.send(None)
```
janezd, line 103: I think it's better if `send` is called from a single function, in this case from `apply`.

There, I'd use `if self.attribute is None`, send `None` and exit.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+            return
+        self.attribute = model[0]
+        self.openContext(data)
+        self.apply()
```
janezd, line 107: `self.apply.now()`, when you have deferred commit.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def apply(self):
```
janezd, line 109: Add decorator `@gui.deferred`.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+        values = self.data.get_column_view(self.attribute)[0]
+        if var.is_discrete:
+            values = values_to_string(values, var.values)
+
+        new_data, new_values = values_from_string(values, self.splitter)
+
+        new_columns = tuple(
+            Orange.data.ContinuousVariable(
+                v,
+                compute_value=OneHotStrings(values, values_from_string(values,
```
janezd, line 120: As I understand, the first argument to `OneHotString` must be a callable.

If I try to use the widget, I always get:  

```python
File "/Users/janez/orange3/Orange/data/table.py", line 300, in get_columns
    shared = col.compute_shared(sourceri)
TypeError: 'numpy.ndarray' object is not callable
```

You need to split `values_from_string` into two functions. The first (which may also be in the widget itself) computes a mapper. The second part is a separate function or a class; it contains that mapper in closure or an attribute. This, second part, is then what you pass to `OntHotString`.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+        var = self.data.domain[self.attribute]
+        values = self.data.get_column_view(self.attribute)[0]
+        if var.is_discrete:
+            values = values_to_string(values, var.values)
+
+        new_data, new_values = values_from_string(values, self.splitter)
+
+        new_columns = tuple(
+            Orange.data.ContinuousVariable(
+                v,
```
janezd, line 119: Variable names must be computed like this:

`var_names = Orange.data.util.get_unique_names(self.data.domain, new_values)`

This will ensure they are unique in case any variables with such names already exist.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+                compute_value=OneHotStrings(values, values_from_string(values,
+                                                                       self.splitter))
+            ) for v in new_values)
+
+        new_domain = Orange.data.Domain(
+            self.data.domain.attributes + new_columns,
+            self.data.domain.class_vars, self.data.domain.metas
+        )
+        extended_data = self.data.transform(new_domain)
+        extended_data.X = np.hstack([self.data.X, new_data])
```
janezd, line 129: This shouldn't be necessary. `self.data.transform(new_domain)` must compute values for the new attributes, otherwise something is wrong with `compute_value`.

It also won't work because tables need to be unlocked.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+        self.Outputs.data.send(extended_data)
+
+
+if __name__ == "__main__":  # pragma: no cover
+    WidgetPreview(OWSplit).run(Table.from_file(
+        "/Users/ajda/Downloads/test.xlsx"))
```
janezd, line 135: Please add a file, and use some kind of relative path so that we can test the widget.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+from AnyQt.QtCore import Qt
+
+from Orange.data import DiscreteVariable, StringVariable
+from Orange.widgets.utils.itemmodels import DomainModel
+from Orange.widgets.utils.widgetpreview import WidgetPreview
+
+foo = ["foo,bar", "foo,baz", "foo", "bar,baz", "baz,foo", "baz", "bar"]
+
+
+def values_from_string(var_list, splitter):
```
janezd, line 17: If there's no reason to have this as a separate function, I'd make it a `@staticmethod` within the class.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+    values = [s.split(splitter) for s in var_list]
+    new_values = sorted(set(val for vals in values for val in vals))
+    mapper = {v: i for i, v in enumerate(new_values)}
+
+    X = np.zeros((len(var_list), len(new_values)))
+    for i, vals in enumerate(values):
+        X[i, list(map(mapper.get, vals))] = 1
```
janezd, line 24: I like `map`, but when you need a list, I believe that a list comprehension is more appropriate, that is `[mapper[val] for val in vals]`, I think.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,135 @@
...
+foo = ["foo,bar", "foo,baz", "foo", "bar,baz", "baz,foo", "baz", "bar"]
+
+
+def values_from_string(var_list, splitter):
+    values = [s.split(splitter) for s in var_list]
+    new_values = sorted(set(val for vals in values for val in vals))
+    mapper = {v: i for i, v in enumerate(new_values)}
+
+    X = np.zeros((len(var_list), len(new_values)))
+    for i, vals in enumerate(values):
```
janezd, line 23: I think this could get slow, but I'm not sure how to optimize it through numpy.

I would think if it could be turned around, so the loop iterates across columns instead of rows. Don't know whether it's possible, though.

janezd (changes requested):
I suddenly became interested in this widget. :)

janezd:
What about a bit of redesign?

Currently:

Screen Shot 2022-05-23 at 16 46 12

Suggestion:

Screen Shot 2022-05-23 at 16 47 43

orangecontrib/prototypes/widgets/owsplit.py
ajdapretnar: Is this not necessary for compute_value?

orangecontrib/prototypes/widgets/owsplit.py
ajdapretnar: But this also considers `compute_value`. Should I create `var_names` separately from `new_columns`?

ajdapretnar:
Q1: Did we agree on keeping a variable comboBox or do we, at any point, wish to have listView (enabling more than a single variable)?

Q2: Do we have a better name for this widget?

wvdvegte:
Just out of curiosity: Will this new widget do the same as the following does in Feature Constructor (with "|" as the delimiter), with the difference that the max. number of classes doesn't need to be predefined? And will it allow different counts of delimiters per record?
Screen Shot 2022-06-06 at 16 56 44

ajdapretnar:
@wvdvegte Kind of, but also not exactly. The position of the value is irrelevant in the proposed widget. So [["foo, bar", "bar, baz", "baz, foo"]] would return [1, 0, 1] for "foo". Also it doesn't count delimiters, it only counts presence of values (0,1) if "foo" in string, for example. We could consider counting how many "foos" are in the string, but I am unsure of the use case that would not benefit more from using bag of words.

wvdvegte:
> We could consider counting how many "foos" are in the string

Thanks for explaining. This seems useful also for some (non-archaeological) data that I've seen recently.
What I meant was, if it can deal with situations where the number of delimited items is different per record (in your case 3 x 2 items). For instance, would it also work with [["foo, bar, bar", "baz", "baz, foo, baz, bar, foo, bar, foo"]]: 3 items - 1 item - 7 items?

ajdapretnar:
Yes, it can. It doesn't care about how many items there are in the string. However, it would not count how many "bars" or "bazes" there are, just 1 if it is present and 0 if not. For proper counts one should use Bag of Words.

ajdapretnar:
On the above topic, @janezd I think we should make the output column a DiscreteVariable instead of ContinuousVariable. Other we duplicate the bow functionality too much.

janezd:
This should depend upon how we use this widget. But from my experience, I'd agree: this is a binary variable.

The alternative would be to have a checkbox to decide whether you'd like to indicate presence or counts. I don't like this idea much, though. For now, I'd go with discrete variables. We can always expand later.

wvdvegte:
> this is a binary variable

"present" / "not present" was also what I was looking for. No frequency count.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,136 @@
...
+def get_string_values(data, var):
+    # turn discrete to string variable
+    column = data.get_column_view(var)[0]
+    if var.is_discrete:
+        return [var.str_val(x) for x in column]
+    return column
+
+
+class SplitColumn:
+    def __init__(self, data, attr, splitter):
```
janezd, line 25: When I first read this, I thought `splitter` would be a function that splits the string.

Perhaps this should be called `delimiter`, not `splitter`.

Same in widget's code.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,136 @@
...
+    splitter = ContextSetting(";")
+    auto_apply = Setting(True)
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        variable_select_box = gui.vBox(self.controlArea, "Variable")
+
+        gui.comboBox(variable_select_box, self, "attribute", label=None,
```
janezd, line 83: If you want no label, give no `label`.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,136 @@
...
+                     orientation=Qt.Horizontal, searchable=True,
+                     callback=self.apply.deferred,
+                     model=DomainModel(valid_types=(StringVariable,
+                                                    DiscreteVariable)))
+        gui.lineEdit(
+            variable_select_box, self, "splitter",
+            orientation=Qt.Horizontal, box=False,
```
janezd, line 90: If you want no box, give no `box`.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,136 @@
...
+from Orange.widgets.widget import OWWidget, Msg, Output, Input
+from AnyQt.QtCore import Qt
+
+from Orange.data import DiscreteVariable, StringVariable
+from Orange.widgets.utils.itemmodels import DomainModel
+from Orange.widgets.utils.widgetpreview import WidgetPreview
+from orangewidget.settings import Setting
+
+
+def get_string_values(data, var):
```
janezd, line 16: This is totally a question of style but: consider changing this function into a `@staticmethod` of class `SplitColumn`. It is use only there, so there's no need to pollute the module's namespace with it.

orangecontrib/prototypes/widgets/owsplit.py
```diff
@@ -0,0 +1,136 @@
+import Orange.data
```
janezd, line 1: Change the order of imports -- first Python's, then numpy/scipy/etc, then Qt's, then `orangewidget`, then `Orange`.

janezd (commented):
Perhaps add some tests that check the actual output of the widget, not just the corner cases (which are of course important, too). :)

orangecontrib/prototypes/widgets/owsplit.py
ajdapretnar: Oof, I am usually quite careful with that. How did I miss it?

ajdapretnar:
TODO: add some general tests

janezd (commented):
I like this widget (including its clean and simple code).

There's a major comment: you will need to implement `__eq__` and `__hash__`.

Another: tests do not check the output (beyond the number of attributes). Please add a tests that actually checks some data.

I would also appreciate separate tests for `OneHotStrings` (and, implicitly `SplitColumn`).

I you need assistance with any of these (or you just really hate to write tests :), please tell me and I'll contribute.

Would we graduate this widget into core once this is complete?

janezd:
If we move the widget to core:

- name could be more informative, but I don't have any good suggestions,
- somebody should write documentation :)

PR #253, opened 2022-07-11
Split: Refactor for discrete values, add tests, rename to "Text to Columns"

janezd:
- Added tests for computation
- Changed treatment of nans for discrete values: 
    - if value is nan, it doesn't mean it *does not* contain particular substrings, so newly created values must be nan.
    - This required too many changes wrt to string variables, so I created a separate class for one hot encoding of discrete values,
    - which are now computed faster (each distinct value is split only once and updated for all appearances),
    - so there is also need for shared computation because it only wastes memory.
- Renamed widget and modules.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
> Counting nans works for discrete but not for string.

I don't get this one. I removed the column with counts of missing values.

Trying this I discovered and fix another bug - if categorical variable was a meta, the widget crashed.

janezd:
> I don't get this one. I removed the column with counts of missing values.

Maybe I do - maybe I just fell into the same trap. Have you been testing Split or Text to Columns? I renamed the widget. :)

janezd:
Not at all. I seldom go to sleep without thinking about it, and @ajdapretnar and I discuss it at least once a week, sometimes even daily.

Seriously, how did you stumble upon this? Yes, it could be merged, I think. I also updated it (now that https://github.com/biolab/orange3/pull/6058 is merged and released). @ajdapretnar, check and merge at will. We can also move it from prototypes; as I recall, it is used for some fields from google forms.

markotoplak:
Failing tests are also not an issue here and we'll fix them later. @erikafuna found a question posted on FB regarding this widget and she asked me which widget that was. Then we searched and found this. Merging. If @ajdapretnar has any comments, let's fix them later.

PR #254, opened 2022-10-07
Owinteractions: new widget and tablemodel

noahnovsak:
##### Description of changes
- `InteractionScorer` relocated and fixed main thread being slowed down by `numpy.unique`. #255
- `RankTableModel` and `ArrayTableModel` added (currently as two separate models).
- `OWInteractions` rewritten as it's own widget, no longer extends `OWCorrelations`.

These changes improve the performance of `OWInteractions` and present some opportunities for `OWCorrelations` and `OWScatterPlot` as well. Currently all three of these widgets can generate models large enough, that `QTableModel` and subsequently the rest of Orange becomes unresponsive. The idea of the new models is to save the underlying data in an array and let _numpy_ do all the heavy lifting. Additionally improvements have been made to the process of adding items into the model in batches (the current process: generate and add items to the queue one by one, take items out one by one, add them to the model one by one).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
@noahnovsak, I merged #255 and rebased this one. Do you remember any reason we didn't merge it a year ago?

noahnovsak:
No, I think it's ok for prototypes... It probably needs some work before merging into the main branch though.

markotoplak:
Hahaha, you got used to different naming schemes! Thank you.

PR #255, opened 2022-10-14
OWInteractions: fix calculation

noahnovsak:
##### Issue
Main thread is bogged down, causing all of Orange to run slowly.

##### Description of changes
Interaction calculation has been fixed to avoid using `np.unique(..., axis=0)`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/interactions.py
+++ b/orangecontrib/prototypes/interactions.py
@@ -41,11 +41,11 @@ def entropy(ar):
     return -np.sum(p * np.log2(p))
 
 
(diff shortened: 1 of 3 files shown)


markotoplak:
This code falls out of int32 range. To guarantee that you are inside, you need to be within A**3  < 2**31 (A**3 is because on the third place could have A**2+A). If A = 10000, that gets us over. I'd say we could do 1000 bins at most, but that is plenty.

PR #256, opened 2022-10-23
Split: whitespace and empty space

janezd:
##### Issue

1. Comma-separated text usually includes a space after comma. When widgets splits by comma, we then have two attributes - those starting with space and those without. Whitespace at the beginning and end is not shown, hence having it will never be intentional, imho. Thus it needs to be removed.
2. If empty strings are considered missing then in case the string is empty, values of all new variables should be shown as missing. Current implementation instead adds an indicator variable for missing strings, which is probably not even used, (r used in case of two consecutive delimiters?) 
3. However, instead of fixing 2, I'd prefer to interpret empty strings differently in this context: they are empty because they do not contain any of the values into which the string is split. I think this makes more sense.

##### Description of changes

Fix 1, fix 2 via 3.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/orangecontrib/prototypes/widgets/tests/test_owsplit.py
+++ b/orangecontrib/prototypes/widgets/tests/test_owsplit.py
@@ -34,7 +34,7 @@ class TestOWSplit(WidgetTest):
             [
                 ["foo,"],
                 ["bar,baz "],
-                ["foo,bar"],
+                ["foo, bar"],
                 [""],
             ]
         )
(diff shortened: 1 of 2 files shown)


orangecontrib/prototypes/widgets/tests/test_owsplit.py
```diff
@@ -34,7 +34,7 @@ def _create_simple_corpus(self) -> None:
             [
                 ["foo,"],
                 ["bar,baz "],
-                ["foo,bar"],
+                ["foo, bar"],
```
janezd, line 37: bar with leading space must be the same as bar without.

janezd:
Now I'm scared. Seriously. I already fixed this in (still to be merged) #253. Now that I see it, I remember.

PR #264, opened 2023-11-22
OWScoringSheet, OWScoringSheetViewer: Add Widgets

ZanMervic:
##### Description of changes
Added two new widgets designed for the Orange Explain addon: OWScoringSheet, OWScoringSheetViewer

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
`fasterrisk`'s dependencies are too strict: 

```
install_requires = \
['numpy>=1.23.3,<2.0.0',
 'pandas==1.5.2',
 'requests>=2.28.1,<3.0.0',
 'scikit-learn==1.2.0']
```

The problems are pandas and scikit-learn. Thus, this is conflicting with Orange. We can either:
1. Convince the maintainer to open them up or
2. Since the code is BSD, we can just put them in our repo (with the proper copyright).

I strongly prefer the first option, but if we are in a rush we'll need to do the second.

PR #266, opened 2023-11-24
OWTextToColumns: add name migration

ajdapretnar:
##### Issue

Old workflows with OWSplit didn't work.

##### Description of changes
Ensure name migration.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
!Note: upon testing with an old workflow, the workflow worked, but the name of the widget remained Split (not Text to Columns as it should be).

markotoplak:
I think keeping the name is fine (users can rename widgets in the workflow anyway, and our workflows do not know whether that was the original name or not). The same thing happens with Formula.

PR #278, opened 2025-03-18
Update owchatgptbase.py

arturborja:
Fix compatibility with OpenAI API models

Corrigindo compatibilidade com modelos no OpenAI API

### Summary
This PR fixes compatibility issues with OpenAI API models by removing unsupported parameters.

### Changes
- Removed `system` role from `messages` to support all models.
- Ensured `temperature` is not explicitly set for models that do not allow custom values.

### Why?
Some models (e.g., `o1-mini`) do not support the `system` role or custom `temperature` values, causing errors.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/orangecontrib/prototypes/widgets/owchatgptbase.py
+++ b/orangecontrib/prototypes/widgets/owchatgptbase.py
@@ -16,8 +16,8 @@ from Orange.widgets.settings import Setting, DomainContextHandler, \
     ContextSetting
 from Orange.widgets.widget import OWWidget, Input, Msg
 
-MODELS = ["gpt-3.5-turbo", "gpt-4",'o3-mini','o1-mini','gpt-4o-mini']
-
+MODELS = ["gpt-3.5-turbo", "gpt-4", "o3-mini", "o1-mini", "gpt-4o-mini"]
+OLD_MODELS = {"gpt-3.5-turbo", "gpt-4"}
 
 def run_gpt(
         api_key: str,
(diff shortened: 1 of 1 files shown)


orangecontrib/prototypes/widgets/owchatgptbase.py
```diff
@@ -16,7 +16,7 @@
     ContextSetting
 from Orange.widgets.widget import OWWidget, Input, Msg
 
-MODELS = ["gpt-3.5-turbo", "gpt-4"]
+MODELS = ["gpt-3.5-turbo", "gpt-4",'o3-mini','o1-mini','gpt-4o-mini']
```
VesnaT, line 19: Please use " instead of ' as elsewhere in the file. 
Besides, spaces between models would be nice.

orangecontrib/prototypes/widgets/owchatgptbase.py
```diff
@@ -27,18 +27,15 @@ def run_gpt(
         prompt_end: str
 ) -> str:
     client = openai.OpenAI(api_key=api_key)
-    enc = tiktoken.encoding_for_model(model)
+    enc = tiktoken.get_encoding("o200k_base")
```
VesnaT, line 30: That is a new encoding that does not appear in the oldest versions of the library, so the library should be limited.

Also, the encoding should not be hard coded, since it is model dependent.

orangecontrib/prototypes/widgets/owchatgptbase.py
```diff
@@ -27,18 +27,15 @@ def run_gpt(
...
     text = enc.decode(enc.encode(text)[:3500])
     content = f"{prompt_start}\n{text}.\n{prompt_end}"
     response = client.chat.completions.create(
         model=model,
-        messages=[
-            {"role": "system", "content": "You are a helpful assistant."},
-            {"role": "user", "content": content},
-        ],
-        temperature=0
+        messages=[{"role": "user", "content": content}],
```
VesnaT, line 36: If some models do not support some parameters, it does not mean those should not be used where possible.

VesnaT (changes requested):

orangecontrib/prototypes/widgets/owchatgptbase.py
arturborja: Double quotes and spaces: Updated MODELS to ["gpt-3.5-turbo", "gpt-4", "o3-mini", "o1-mini", "gpt-4o-mini"] with " and spaces for consistency and readability.

orangecontrib/prototypes/widgets/owchatgptbase.py
arturborja: Parameters: Kept temperature=0 for old models where it’s supported, omitted it for new ones that may not, using it where possible.

orangecontrib/prototypes/widgets/owchatgptbase.py
arturborja: Encoding and library: Replaced hardcoded "o200k_base" with tiktoken.encoding_for_model(model) to match the encoding available for each model. Suggest tiktoken>=0.7.0 in requirements.txt for new encoding support.

arturborja:
@VesnaT I’m looking forward to you accepting my pull request 
Please

VesnaT:
Thank you for your contribution. However, we are afraid that the solution is outdated, and therefore we are closing this pull request.

Issue #279, opened 2025-08-14
Text to columns: suggested enhancements


wvdvegte:
The Text-to-columns widget is very useful and doesn't seem to be buggy (so I don't know why it is still in Prototypes) but I think it could benefit from the following improvements:

1. The columns created based on the separated text items are categorical by default. Since the categories are '0' and '1' it would be nice to make them readily available for summation by making them numeric by default. For instance, if the categories are languages and you would like to define a variable that expresses whether at least one of the languages English, French and Slovenian is included, you could use the expression `English + French + Slovenian > 0` in Formula. With categories, you need to write out each reference to a language as `int(Language)` which is cumbersome.
2. It would be nice to have the option to convert text from multiple text items to (new) columns. The current widget needs to be inserted repeatedly to process these. For instance, the [IMDb movies dataset](https://www.kaggle.com/datasets/thedevastator/imdb-movie-and-crew-data) has text items for genres, languages, countries etc. that would all lend themselves for conversion to multiple new columns. Since these all use the same separator (which is likely also true for similar datasets), it would not be necessary to add options for defining separators for each text item separately.

wvdvegte:
Thanks for picking up the first suggestion - encountering this little problem was what triggered me to post this.
Regarding the second point - the mess: I guess in practice it will indeed be used in cases where "multiple"  is two or three, for text that doesn't cover to many different variables. But it would be nice not to have to "chain" the widget in those cases.
The other suggestion causes long column names, which are often difficult to handle - especially since, in many cases, you have to manually widen each column to see the second, distinctive part.

janezd:
To be discussed. Please remind me tomorrow if I forget about this add-on.

ajdapretnar:
1. Widget moved.
2. Renamed to Split.
3. Numeric output added.
4. We decided against multiple selections because this would complicate the widget quite a bit. We might change our minds at some point in the future.

PR #280, opened 2025-08-15
Text To Columns: Allow setting output format

janezd:
##### Issue

Resolves part of #279: output formats.

In addition to what @wvdvegte wrote, the current output (0 and 1) is ambiguous: it looks like number, but these are "categories".

##### Description of changes

Adds output format which can be No / Yes, 0 / 1 (as number) or counts.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
I don't like this GUI. It looks to "technical". Suggestions welcome.

Screenshot 2025-08-15 at 16 05 32

PR #282, opened 2025-08-15
Quick Select tests

janezd:
##### Issue

Quick Select cannot be matured from prototypes because it has no tests.

##### Description of changes

Quick Select can be matured from prototypes because it has tests.

It now also has lazy signals.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
-> Educational

PR #286, opened 2026-06-17
Add Neighbor Joining widget

drejcbevk:
##### Issue
Adds a prototype Neighbor Joining widget.

##### Description of changes
Added files:

* `neighbor_joining.py`
  Implements the core algorithm, `TreeNode` structure, leaf ordering helpers and Newick export.

* `neighbor_joining_adapter.py`
  Converts the `TreeNode` structure into Orange's `hierarchical.Tree` structure.

* `dendrogram_nj.py`
  Custom dendrogram component, supports non-ultrametric trees by preserving leaf heights and branch lengths.

* `owneighborjoining.py`
  Adds the Orange widget for visualizing, pruning, selecting, and outputting clusters.

* `test_neighbor_joining.py`
  Tests for core algorithm and tree adapter.

* `test_owneighborjoining.py`
  Widget tests for input handling, error states, tree construction, and outputs.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -468,48 +384,62 @@ def axis_view(orientation):
...
+                    return
+                if data.domain.has_continuous_attributes():
+                    matrix = distance.Euclidean(data, axis=1, impute=True)
+                elif data.domain.has_discrete_attributes():
+                    matrix = distance.Hamming(data, axis=1, impute=True)
+                else:
+                    self.Error.no_numeric_features()
+                    return
+                self._set_valid_matrix(matrix)
+            except Exception as e:
```
markotoplak, line 420: Catching all exceptions and displaying them as errors is usually a bad practice. If something is not supposed to crash and you have relatively good control, then it is ever beneficial if an actual exception happens: users get notified and could report it.

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -468,48 +384,62 @@ def axis_view(orientation):
...
-                self.matrix = matrix
+            try:
+                row_items = getattr(matrix, "row_items", None)
+                if isinstance(row_items, Orange.data.Table) \
+                        and not self._has_input_features(row_items):
+                    self.Error.no_numeric_features()
+                    return
+                self._set_valid_matrix(matrix)
+            except Exception as e:
+                self.Error.distance_computation_error(error=str(e))
```
markotoplak, line 399: Catching all exceptions and displaying them as errors is usually a bad practice. If something is not supposed to crash and you have relatively good control, then it is ever beneficial if an actual exception happens: users get notified and could report it.

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -587,43 +519,54 @@ def _set_items(self, items, axis=1):
...
-                self._set_displayed_root(tree)
+                self.root = tree
+                self.top_axis.setRange(tree.value.height, 0.0)
+                self.bottom_axis.setRange(tree.value.height, 0.0)
+
+                if self.pruning:
+                    self._set_displayed_root(prune(tree, level=self.max_depth))
+                else:
+                    self._set_displayed_root(tree)
+            except Exception as ex:
```
markotoplak, line 557: Ditto.

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -641,19 +584,31 @@ def _update_labels(self):
...
+                    # Empty strings let the dendrogram still show subset bolding.
                     labels = [""] * len(indices)
                 else:
                     labels = []
             elif self.annotation == "Enumeration":
                 labels = [str(i+1) for i in indices]
             elif self.annotation == "Name":
-                attr = self.matrix.row_items.domain.attributes
-                labels = [str(attr[i]) for i in indices]
+                row_items = self.matrix.row_items
```
markotoplak, line 594: Why did the logic change here, for annotation? Why were changes here needed at all - in the and, the annotatio are shown similarly.

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -663,68 +618,51 @@ def _update_labels(self):
...
+                        ", ".join(index_to_label[i] for i in leaf_indices)
+                    )
 
-        self.label_model[:] = labels
-        self.label_model.set_subset(set() if self.pruning else self.subset_rows)
-        self.labels.setMinimumWidth(1 if labels else -1)
+                labels = new_labels
 
-        if not self.pruning and self.color_by is not None:
+        # Compute per-leaf colors (for the small rectangle) and subset bold mask
```
markotoplak, line 635: For me it is not obvious why these parts had to be changed.

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -925,9 +865,34 @@ def _set_cut_line_visible(self, visible):
 
     def select_top_n(self, n):
         root = self._displayed_root
-        if root:
-            clusters = top_clusters(root, n)
-            self.dendrogram.set_selected_clusters(clusters)
+        if not root:
```
markotoplak, line 868: This function is an example of a function that has good arguments why it needs to be different to NJ

orangecontrib/prototypes/widgets/owneighborjoining.py
```diff
@@ -468,48 +384,62 @@ def axis_view(orientation):
...
+                if isinstance(row_items, Orange.data.Table) \
+                        and not self._has_input_features(row_items):
+                    self.Error.no_numeric_features()
+                    return
+                self._set_valid_matrix(matrix)
+            except Exception as e:
+                self.Error.distance_computation_error(error=str(e))
+
+    @Inputs.data
+    def set_data(self, data):
```
markotoplak, line 402: Adding Data should be handled carefully. I'd prefer not to have it, because it contains too many automatic choices that are not transparent.

At minimum it should be transparent. So, if it made a choice, that choice would have to be reported to the user in some kind of an info box. Then, it assumes a specific axis, which is a big assumption, and it does not standardize the data. In such cases the proper thing would be to have a combo where the user would select the axis (rows/colums) and a normalization/standardization check box, because these are important parameters. But if you added all these, then there is no gain compared to using explicit **Distance**.

I am commenting this from the point of view of Orange UI design principles.

